PR #109, opened 2017-08-07
PR: Add bootstrap envs

fpliger:


anaconda_project/internal/cli/environment_commands.py
```diff
@@ -226,3 +230,44 @@ def main_update(args):
 def main_unlock(args):
     """Unlock dependency versions and return exit status code."""
     return unlock(args.directory, args.name)
+
+
+def running_in_bootstrap_env(project_dir):
```
goanpeca: Minor nitpick, but if this is returning True or False a name `is_running_in_bootstrap_env` is better :-p

anaconda_project/internal/cli/run.py
```diff
@@ -37,29 +38,36 @@ def run_command(project_dir, ui_mode, conda_environment, command_name, extra_com
...
-              project_dir,
-              file=sys.stderr)
-    else:
-        try:
-            result.command_exec_info.execvpe()
-        except OSError as e:
-            print("Failed to execute '%s': %s" % (" ".join(result.command_exec_info.args), e.strerror), file=sys.stderr)
+        if result.failed:
+            # errors were printed already
+            return
```
goanpeca: Why do we need to return? I feel having returns in the middle of a routine/procedure/method/func is weird. This is more like a procedure than a function, so it does not return anything, it just does stuff. In this case, a `pass` feels more correct, besides if if entered the if, there is no code that "might get executed" after the `else`

anaconda_project/project.py
```diff
@@ -842,6 +844,10 @@ def _update_conda_env_requirements(self, requirements, problems, project_file):
         if _fatal_problem(problems):
             return
 
+        if 'bootstrap-env' in self.env_specs:
```
goanpeca: This check feels a but manual, and is repeated on several places, maybe we could have a helper method?

Why not use `def has_bootstrap_env_spec(self):` ?

anaconda_project/project.py
```diff
@@ -1573,3 +1585,18 @@ def use_changes_without_saving(self):
         """
         self.project_file.use_changes_without_saving()
         self.lock_file.use_changes_without_saving()
+
+    @property
+    def bootstrap_env_prefix(self):
```
goanpeca: This does feel like a property

anaconda_project/project.py
```diff
@@ -1573,3 +1585,18 @@ def use_changes_without_saving(self):
...
+        """Fullpath to bootstrap environment prefix."""
+        return join(self._directory_path, 'envs', 'bootstrap-env')
+
+    @property
+    def is_running_in_bootstrap_env(self):
```
goanpeca: This does **not** feel like a property, this is answering a question... like a function.

anaconda_project/project.py
```diff
@@ -1573,3 +1585,18 @@ def use_changes_without_saving(self):
...
+        """Return True if anaconda-project is running inside a project bootstrap env False otherwise."""
+        return os.environ['CONDA_PREFIX'] == self.bootstrap_env_prefix
+
+    @property
+    def has_bootstrap_env_spec(self):
```
goanpeca: This does **not** feel like a property, this is answering a question... like a function.

`bootstrap_env_spec` could return the spec or None but `has_bootstrap_env_spec` feels like a function

anaconda_project/internal/cli/environment_commands.py
fpliger: +1

anaconda_project/internal/cli/run.py
fpliger: This code is not part of the PR (it has just been moved, so the `return` stmnt was already there). Returning in the middle of function-like object does not bother me.. I know returning the result of a function at the end of a function (forcing this "standard" pattern maybe storing results in a variable that you carry until the end) is what have been thought for decades but I think it is ok to return prematuraly. Especially in Python and if this kind of helps understand the overall function behavior.

I'd be -0 about changing it.

anaconda_project/project.py
fpliger: right.. especially since it is already there.. 😄

anaconda_project/project.py
fpliger: mmmm.... kind of disagree.. A property __is__ a function... just a special one. This one happens to behave like a flag. There is not rule that says properties cannot behave like status checks variables/functions, right?

anaconda_project/project.py
fpliger: Maybe... yeah. Need to check the code again to reply on this one.

anaconda_project/project.py
goanpeca: >A property is a function... just a special one. 

I know the internals, I am talking about how properties are used, they are basically the state of something so naming them in ways that they feel like a function is what I am arguing here.

```

class Fabio():

    @property
    def age(self):
        return datetime.now() - self.date_of_birth

    def mood(self):
        return self._mood

    def is_angry(self):
        return self._mood == self.MOOD_ANGRY
 ```

fpliger: meh... yeah, I get your point. I just feel like it's more bikeshedding than really adding value. I can do the change anyway :)

goanpeca: You are the boss ;-) I make the nitpicks, you make the calls

goanpeca: But we do need the rebase/merge with master though!

fpliger: Yup.. Will do it but let's leave it out of the dev build today

PR #116, opened 2017-09-19
Config envs path

fpliger:
This enables customization of envs_path through PROJECT_ENVS_PATH env variable

### DIFF
--- a/anaconda_project/env_spec.py
+++ b/anaconda_project/env_spec.py
@@ -307,7 +307,7 @@ class EnvSpec(object):
 
     def path(self, project_dir):
         """The filesystem path to the default conda env containing our packages."""
-        envs_path = os.environ.get('PROJECT_ENVS_PATH', os.path.join(project_dir, "envs"))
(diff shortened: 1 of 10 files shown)


anaconda_project/env_spec.py
```diff
@@ -307,7 +307,9 @@ def inherit_from_names(self):
 
     def path(self, project_dir):
         """The filesystem path to the default conda env containing our packages."""
-        return os.path.join(project_dir, "envs", self.name)
+        envs_path = os.environ.get('PROJECT_ENVS_PATH', os.path.join(project_dir, "envs"))
```
goanpeca: @fpliger should we use a more specific variable? Like:

`ANACONDA_PROJECT_PATH` `ANACONDA_PROJECT_ENVS`?

It is nice that environment variables are specific and not generic, I feel `PROJECT_ENVS_PATH` is too generic

fpliger: Agree.

PR #118, opened 2017-09-21
[WIP] change envs path 2

fpliger:


fpliger:
@havocp do you have any hint about why this is failing? Something (some dependency?!) seems to have broken tests but I can't figure out what...

PR #137, opened 2017-11-14
Ignore all ipynb_checkpoints directories

mcg1969:
During an attempt to create a project from an AENv4 project I noticed that it scanned `.ipynb_checkpoints` directories that were not in the project root. It seems safe to avoid doing this.


mcg1969:
The Travis failure is for the mac build, which stalled at `anaconda_project/test/test_prepare.py::test_prepare_empty_directory`.

PR #144, opened 2017-12-13
Remove --no-deps from pip install command

koverholt:
Removing `--no-deps` from pip install command to match up with user expectations around pip/conda and resolving dependencies.

cc @fpliger

### DIFF
--- a/anaconda_project/internal/pip_api.py
+++ b/anaconda_project/internal/pip_api.py
@@ -64,9 +64,7 @@ def install(prefix, pkgs=None):
         raise TypeError('must specify a list of one or more packages to install into existing environment, not %r' %
                         pkgs)
 
-    # --no-deps is because we don't want to pull in pip versions of
-    # everything that conda has.
-    args = ['install', '--quiet', '--no-deps']
+    args = ['install', '--quiet']
     args.extend(pkgs)
 
     return _call_pip(prefix, extra_args=args)


fpliger:
@koverholt what will happen if some of the deps are already installed by conda? and what happens if you conda update or conda remove later? Maybe pip is smart enough and will avoid installing packages already installed by conda but worth checking that actually behavior matches expectations. (it'd be also good to have tests covering those cases)

Another thing to consider is that if you don’t have a compiler, there are limits to what pip can do (i.e. no C extensions)... 

In my mind, the best option would be to bubble this as an option argument if we want to change the default behavior (so at lease we have a backup plan).

danielfrg:
bump @fpliger pls do this

cpbotha:
`--no-deps` should definitely be removed. My reasons for saying this:

- using pip inside of a conda environment is a use-case which occurs often; less so the other way round. `anaconda-project` first resolves the conda environment, and then does the pip installs.
- with `--no-deps` and more than one or two pip packages, you are bound to have a broken environment. Put differently: people are going to run into broken environments much more often than not.
- because pip and conda work in the same site-packages, pip generally recognizes conda-installed packages.
- I've just double-checked with the latest version of conda as of today: `conda env` runs `pip install` WITH the resolving of dependencies. Just to ram home that point: I am working on a conda environment with additional pip packages at this moment which `conda env` handles with aplomb, but with which `anaconda-project` builds an unusable environment.

I hope that helped. Please merge this PR as soon as you can.

mcg1969:
@cpbotha, you make a great case. @fpliger, I'm personally convinced. Any thoughts?

mcg1969:
I would say that I don't really think it's anaconda-project's job to police conda's behavior here. I think that if we simply codify that anaconda-project executes conda first and pip second, we're fine.

fpliger:
+1 on my side as well... (sorry for the delay :) )

PR #167, opened 2018-12-08
Make tests pass with latest pip/conda/tornado

mcg1969:
I'm not seeing any downside to replacing `pip list <args>` with `pip freeze`. It greatly simplifies the pip implementation.

But more importantly, the tests were just not working with later versions of conda and pip.


anaconda_project/internal/pip_api.py
```diff
@@ -87,27 +87,10 @@ def installed(prefix):
     if not os.path.isdir(prefix):
         return dict()
 
-    # In pip 9, there's a big ugly deprecation warning by default if
```
mcg1969: By replacing `pip list` with `pip freeze` we avoid a ton of painful logic figuring out what `--format` keyword argument to use. In fact, the current version of this code is broken with newer versions of pip, because `--format=legacy` is gone.

anaconda_project/internal/conda_api.py
```diff
@@ -21,6 +21,8 @@
 from anaconda_project.internal.directory_contains import subdirectory_relative_to_directory
 from anaconda_project.internal.py2_compat import is_string
 
+CONDA_EXE = os.environ.get("CONDA_EXE", "conda")
```
mcg1969: This is necessary for versions of conda that do not put conda on the path by default, choosing instead to use a shell script.

mcg1969:
@goanpeca @fpliger do you think it's possible to get this reviewed and merged in? This is basically a "get tests passing again" PR, though with fixes required to get it working with later versions of conda and pip (versions that presumably will end up in future versions of AE).

There is more I want to do here, for sure. But making sure we have functioning tests is a necessary thing, even if I don't.

mcg1969:
That last commit allows tests to pass on Windows.

I personally would like to refactor these tests considerably. In particular, I'd like all _functional_ testing to be performed _on the conda package_ within conda build itself. Thankfully this seems entirely doable, and I have a conda recipe that accomplishes this.

Then the source verification testing---flake8, yapf, pep257, copyright notices, etc---would go into a _separate_ Travis task, run only once. There would still be an expectation that the developer commits any of these textual changes themselves.

mcg1969:
But for now—this at least gets the tests working again, and the code working with the latest conda and pip. Those additional refactors can be done in a separate PR.

mcg1969:
I didn't mean to close this. I did however fix the tornado issue :-)

mcg1969:
The only reason this isn't passing is because conda is taking so long to install `anaconda` in some of the tests. I know the conda team is working on this; I'll see if there's a way to reduce test time here.

mcg1969:
I've reduced test time in a three ways:
- Monkeypatching the default environment spec for empty projects to contain `python 3.7` instead of `anaconda`, which speeds up solve time considerably.
- Adding some version pins to a couple of tests that were creating projects with explicit package lists.
- Bumping up some of the pinned versions of `bokeh`, `conda`, and `python`, so they can be found in `pkgs/main` instead of `pkgs/free`. This allows us to remove `pkgs/free` from our default channel list, if we choose. Unfortunately, this breaks the `msys2` test, so we can't do this in CI, but it's useful for local testing (and a good strategy for conda use in general).

fpliger:
@mcg1969 The change LGTM.. we can merge as soon as we get CI happy again :)

mcg1969:
Yep, I'll fix things

mcg1969:
It's going to pass this time.

mcg1969:
@fpliger ready to merge! I'll do a second PR streamlining the tests and making better use of conda-build once this is merged. No functional changes on that one either.

PR #170, opened 2018-12-13
CI simplification; use conda build

mcg1969:
Do functional testing within conda build. Separate formatting tests into a separate CI.


mcg1969:
@fpliger @goanpeca FYI, if you happen to be following along, this is almost done. I'm testing the AppVeyor builds on my fork, and there remains a couple of Python 2.7 Windows failures related to line endings.

mcg1969:
OK! So this passes and simplifies the build process considerably. 
- The CI specs `appveyor.yml` and `.travis.yml` have been simplified considerably.
- All functional testing is performed by `conda build`
- AppVeyor passes for 2.7, 3.5, and 3.6, so we can (and, IMO, should) add it back

PR #177, opened 2019-01-06
Wait for subprocess and exit with same return code

### LINKED ISSUE #154: anaconda-project switching between prompts for commands on Windows
When running a command that displays output, the prompt is returned before the command is fully executed. This happens in Git Bash, Hyper, and cmd.exe on Windows. Although this example is trivial, it leads to larger issues with commands that use interactive prompts

#### anaconda-project.yml

```
name: test

commands:
  pyi:
    windows: python -i 
    env_spec: foo
  pyp:
    windows: python -c "import time;time.sleep(5);print('\nexample\noutput')"
    env_spec: foo
env_specs:
  foo:
    packages:
      - python >=3.6,<3.7
```

Running `anaconda-project pyp`:
```

C:\Users\user\Desktop\test>anaconda-project run pyp

C:\Users\user\Desktop\test>
example
output
```

The expected result should be printing before the prompt gets returned.

Running `anaconda-project pyi` switches between the normal prompt and the python prompt

```
C:\Users\user\Desktop\test>anaconda-project run py

C:\Users\user\Desktop\test>Python 3.6.5 |Anaconda, Inc.| (default, Mar 29 2018, 13:32:41) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import this
'import' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\user\Desktop\test>anaconda-project run pyp
  File "<stdin>", line 1
    anaconda-project run pyp
                       ^
SyntaxError: invalid syntax
>>>
C:\Users\user\Desktop\test>
>>>
C:\Users\user\Desktop\test>
>>>
C:\Users\user\Desktop\test>
>>>
C:\Users\user\Desktop\test>
>>>
```

This was all done by creating the `anaconda-project.yml` file and running `anaconda-project prepare`.

mcg1969:
Fixes #154.
Rebases #158; original text from there:

-------

Not sure how this made it through this long, but the current behavior is a pretty awful experience: random interleaved outputs (#154), broken usage when scripted, and failures not being caught.

Can work up a proper test if necessary... basically a project file like:
```yaml
commands:
  a:
    windows: timeout 1 && echo a
  b:
    windows: echo b
```
then
```bat
anaconda-project run a
anaconda-project run b
```
would yield
```
b # immediately
a # after a second
```

and should yield

```
a # after a second
b # immediately
```

### DIFF
--- a/anaconda_project/project_commands.py
+++ b/anaconda_project/project_commands.py
@@ -262,8 +262,7 @@ class CommandExecInfo(object):
                 # the Windows API calls. We need to do that, rather
                 # than calling os.execvpe which doesn't let us set those
                 # flags. So we spawn the child and then exit.
-                self.popen()
-                sys.exit(0)
+                sys.exit(self.popen().wait())
             else:
                 # this is all shell=True does on unix
                 args = ['/bin/sh', '-c'] + args
--- a/anaconda_project/test/test_project_commands.py
+++ b/anaconda_project/test/test_project_commands.py
@@ -49,6 +49,12 @@ def test_execvpe_with_shell_on_windows(monkeypatch):
         executed['cwd'] = cwd
         executed['shell'] = shell
 
+        class MockProcess(object):
+            def wait(self):
+                return 0
+
+        return MockProcess()
+
     monkeypatch.setattr('subprocess.Popen', mock_popen)
 
     info = CommandExecInfo(cwd='/somewhere', args=['foo bar'], shell=True, env=dict(FOO='bar'))


bollwyvl:
I probably can't update on this, but it looks like this might do the trick:
```py
def mock_popen(args, env, cwd, shell):
    executed['args'] = args
    executed['env'] = env
    executed['cwd'] = cwd
    executed['shell'] = shell

    class MockProcess(object):
        def wait(self):
            return 0

    return MockProcess()
```

mcg1969:
Thanks Nick!

bollwyvl:
Godspeed! Thanks for picking up all these pieces: I've been using project a lot and this is really the only thing I've had to maintain a patch for.

mcg1969:
That's great to here. It has several internal fans as well and it was high time it got the love it deserves.

bollwyvl:
Hooray!

PR #182, opened 2019-02-02
Fix small typo so default var renders correctly

cpbotha:
This fixes a minor typo so that the yaml source code renders directly. See screenshot below for current version on the left, the fixed version on the right:

image


mcg1969:
Thank you!

Issue #194, opened 2019-05-21
Allow "dependencies" as an alias for "packages"?


jbednar:
Am anaconda-project.yml file includes nearly the same information as in a Conda environment.yml file, while also adding other metadata like commands and datasets to run.  I personally find an anaconda-project.yml file to be a useful way to capture all of that information, but sometimes I only want to create the conda environment, without using any of the other info. Given that there are far more conda users than anaconda-project users, it's useful to be able to manually create a conda environment with conda commands rather than the ones specific to anaconda-project. 

Julia Signell has made examples of achieving this, such as for the [bay_trimesh](https://github.com/pyviz-topics/examples/tree/master/bay_trimesh) PyViz demo. Specifically, she used YAML references to create a "packages" field in the .yml file that refers to the list declared in the "dependencies" field:

```
dependencies: &deps
  - python=3.6
  - notebook ==5.7.8

packages: *deps
```

Having done so, the .yml file now works with conda as if it were an environment.yml file:

```
conda env create --file anaconda-project.yml
```

That's great and already works, but it's mysterious to readers, annoying to have to do, and currently generates a warning about anaconda-project not understanding the "packages" field.  It would be nice if anaconda-project could follow conda syntax and accept `dependencies` as an alias for `packages`, and indeed to suggest that people generally use `dependencies` so that their files will just automatically work with conda as-is.

This is another example of unifying the .yml syntax with what conda accepts more generally, such as already proposed in https://github.com/Anaconda-Platform/anaconda-project/issues/193.

mcg1969:
Addressed by #200

PR #200, opened 2019-06-06
Allow packages key

cjmartian:
Allows for the use of 'packages' as an alias for 'dependencies' as requested here:
https://github.com/Anaconda-Platform/anaconda-project/issues/194

### DIFF
--- a/anaconda_project/env_spec.py
+++ b/anaconda_project/env_spec.py
@@ -438,7 +438,7 @@ def _load_environment_yml(filename):
     # keeping the new project file if it's messed up.
     #
     # However we do try to avoid crashing on None or type errors here.
-    print(type(yaml))
+    
     raw_dependencies = yaml.get('dependencies', [])
     if not isinstance(raw_dependencies, list) or len(raw_dependencies) == 0:
         raw_dependencies = yaml.get('packages', [])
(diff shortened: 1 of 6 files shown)


cjmartian:
Just ran `python scripts/run_tests.py` locally on my mac:
anaconda_project/requirements_registry/providers/test/test_conda_env.py::test_prepare_and_unprepare_project_scoped_env PASSED
`
python --version
Python 3.7.3
`

The travis-ci failure is on mac build target 3.7

mcg1969:
@cjmartian, sorry this took so long to look at but I have some bad news—this isn't actually what was asked for in #194. What that issue is requesting is for the syntax for `anaconda-project.yml` itself to be altered, so that `dependencies:` can be freely used wherever `packages:` is currently expected. To see what this means, take a look at this file, referred to in #194:

https://github.com/pyviz-topics/examples/blob/master/bay_trimesh/anaconda-project.yml

Note the use of `dependencies:`, but also this `&deps`/`*deps` trick.

cjmartian:
This branch should have been named `allow-dependencies-key` sorry about that.

I tested that I can still run `anaconda-project run` with an `anaconda-project.yml` from the examples directory that has the key `dependencies` rather than `packages` and it seems to work.

PR #209, opened 2019-06-29
Slight test change to ensure conda 3.7 passes

mcg1969:


mcg1969:
To be clear—the original tests _should_ pass with conda 3.7. But due to [this issue](https://github.com/conda/conda/issues/8848) it does not.

mcg1969:
And, it looks like that issue is a wontfix, so this test change will be permanent. Oh well.

PR #210, opened 2019-07-02
Add support for old typo

mcg1969:
Some dependent projects depended on the misspelled class name `ArgsTrasformerTemplate`. Adding that as a synonym back in.


mcg1969:
Conda 4.7 is breaking things I'm afraid. I think it's due to dropping the free channel.

Issue #189, opened 2019-03-25
Prepare stops on first env that already exists


AlbertDeFusco:
The scenario is that I have prepared an `anaconda-project` directory. I then want to add another env by manually editing the `anaconda-project.yml` file. The behavior is different depending on where the second env is added.

## Preparation
Here is the base `anaconda-project.yml` file.

```yaml
name: multiple envs

packages: []
channels: []

env_specs:
  first_env:
    packages:
      - python=3.6
      - requests
```

Create the `envs` directory by running `anaconda-project prepare`. Here's the bottom of the output

```
#
# To activate this environment, use
#
#     $ conda activate /Users/adefusco/Desktop/my_project/envs/first_env
#
# To deactivate an active environment, use
#
#     $ conda deactivate

The project is ready to run commands.
Use `anaconda-project list-commands` to see what's available.
```

## Second env last will never get created

If the new environment is written *AFTER* the first, it will not get created. `anaconda-project prepare` ignores the second env.

```yaml
name: multiple envs

packages: []
channels: []

env_specs:
  first_env:
    packages:
      - python=3.6
      - requests
  second_env:
    packages:
      - python=3.7
      - requests
```

The output of `anaconda-project prepare` indicates that no actions were taken

```
(base) adefusco@0235-adefusco:my_project>anaconda-project prepare
The project is ready to run commands.
Use `anaconda-project list-commands` to see what's available.
```

## Second env first will get created
If the second environment is listed *BEFORE* the one that was already created it will be prepared.

```yaml
name: multiple envs

packages: []
channels: []

env_specs:
  second_env:
    packages:
      - python=3.7
      - requests
  first_env:
    packages:
      - python=3.6
      - requests

```

Here is the output of `anaconda-project prepare`

```
#
# To activate this environment, use
#
#     $ conda activate /Users/adefusco/Desktop/my_project/envs/second_env
#
# To deactivate an active environment, use
#
#     $ conda deactivate

The project is ready to run commands.
Use `anaconda-project list-commands` to see what's available.
```

zdog234:
This isn't a great answer, but if you want to prepare `second_env`, you can use `anaconda-project prepare --env-spec second_env`

I do think that the behavior of only creating the first one is a bit confusing. Would the `anaconda-project` maintainers be up for a PR to change this behavior? (Or to implement a `--all`/`--allenvs` option).

cjmartian:
Implementing this in #213

[...]

PR #213, opened 2019-07-09
prepare --all

cjmartian:
This allows for the `--all` flag for prepare that will iteratively create each environment in a project yaml, given that there is more than one.


cjmartian:
This is for #189

PR #215, opened 2019-07-09
Subdirectories in projects

AlbertDeFusco:
Like git anaconda-project commands search parent directories until an `anaconda-project.yml` file is found.

The only exception is that `anaconda-project init` will *always* create a new anaconda-project.yml file in the working directory. Again, this is just like git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,8 @@
 .sw[nop]
 *.tmp
 *.coverage.*
-/.idea
-/.vscode
+.idea/
+.vscode/
 
 # Compiled source #
 ###################
--- a/anaconda_project/internal/cli/test/test_bundle.py
+++ b/anaconda_project/internal/cli/test/test_bundle.py
@@ -43,12 +43,14 @@ def test_archive_command_on_simple_project(capsys):
                             "  Consider using the 'anaconda-project lock' command to lock the project.")
 
         out, err = capsys.readouterr()
-        assert ('  added %s\n  added %s\n%s\nCreated project archive %s\n' %
-                (os.path.join("some_name", DEFAULT_PROJECT_FILENAME), os.path.join(
-                    "some_name", "foo.py"), unlocked_warning, archivefile)) == out
+        assert ('  added %s\n  added %s\n  added %s\n%s\nCreated project archive %s\n' %
+                (os.path.join('some_name', '.projectignore'),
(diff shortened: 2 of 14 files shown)


mcg1969:
Rebased this off of master, now that tests are passing again.

mcg1969:
I'm seeing this error:
```
E       AssertionError: assert '  added some...s8r/foo.zip\n' == '  added some_...s8r/foo.zip\n'
E         +   added some_name/.projectignore
E             added some_name/anaconda-project.yml
E             added some_name/foo.py
E           Warning: env specs are not locked, which means they may not work consistently for others or when deployed.
E             Consider using the 'anaconda-project lock' command to lock the project.
E           Created project archive /Users/mgrant/Repos/anaconda-project/build/tmp/test-zl5rss8r/foo.zip
```
It looks like the assertion is not expecting the lock warning.

mcg1969:
@AlbertDeFusco rebased this one more time now that i've got the tests all sorted and the other PR's merged in. Please work from this when you get a chance...

mcg1969:
Ah, I see—what seems to be happening is that `.projectignore` is now being added more consistently to the test projects by default, and a lot of the tests aren't taking that into account. I think that's the right behavior, so it is the tests that need adjusting

AlbertDeFusco:
That's something I can check on when I get back to this. I did make a little change to make sure the file gets written, but there are several use cases with anaconda-project now where that file does not get written. I think one of them is if I write my own anaconda-project.yml and run anaconda-project prepare it does not write the ignore file.

AlbertDeFusco:
I'm going through the tests and finding that I need to add `.projectignore` and remove `anaconda-project-local.yml`. Not sure if my changes are causing this.

mcg1969:
Seems it has to be so. We've merged quite a few PRs with fully passing tests lately that don't touch this functionality.

AlbertDeFusco:
Ah, ha! The `anaconda-project-local.yml` should *not* be present because the projectignore file says it doesn't belong there. So by ensuring that the projectignore file is created more consistently the project-local file is not correctly excluded from the archives. I'll continue to update tests and push more commits.

https://github.com/Anaconda-Platform/anaconda-project/blob/3804895ffcea4f2a88747b245fb493a1c0d777bc/anaconda_project/project_ops.py#L38

mcg1969:
I see lots of green above! Are you ready to merge this?

PR #224, opened 2019-08-15
add user_fields field

cjmartian:
This adds support for a `user_fields` field that will stop warnings from happening and will allow the user to define arbitrary fields in their project.yaml as requested here: https://github.com/Anaconda-Platform/anaconda-project/issues/221

PR #228, opened 2019-10-06
User fields

mcg1969:
Rebased version of #224


mcg1969:
@cjmartian FYI. Pulled your functional commits in, but left out all of the adjustments to CI.

PR #230, opened 2019-10-06
Correct failing tests on OSX / Python 3.7

mcg1969:
Trying to figure out why Mac OSX / Python 3.7 fails spectacularly in Travis but not locally


mcg1969:
Looks like it was due to the custom `TemporaryDirectory` equivalent we were using. Opted for the native version for Python 3 and `backports.tempfile.TemporaryDirectory` for Python 2.

PR #232, opened 2019-10-16
Autogenerate default env

AlbertDeFusco:
This allows the simplest anaconda-project.yml file to be utilized without an `env_specs` key.

```yaml
name: My project

packages:
  - python=3.7
  - notebook
channels: []
```

on `anaconda-project prepare` it will infer that the name of the `env_spec` is `default`.

### DIFF
--- a/anaconda_project/internal/cli/test/test_project_load.py
+++ b/anaconda_project/internal/cli/test/test_project_load.py
@@ -10,7 +10,6 @@ from __future__ import absolute_import, print_function
 import os
 import sys
 
-from anaconda_project.project import Project
 from anaconda_project.project_file import DEFAULT_PROJECT_FILENAME
 from anaconda_project.internal.cli.project_load import load_project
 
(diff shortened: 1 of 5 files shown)


AlbertDeFusco:
This also works on `environment.yml`-like files. This PR could be extended to just assume the `environment.yml` file is the `anaconda-project.yml`. Anyone have strong opinions here?

```yaml
name: no_envs
dependencies:
  - python=3.7
  - requests
```

jbednar:
I'm not sure of the precise thing you are proposing, but it sounds in line with my own opinion that there shouldn't be any major distinctions between an environment.yml file and an anaconda-project.yml file; as far as environments are concerned they should be made to be interchangeable.

mcg1969:
I am all for this. I would definitely support interpreting `environment.yml` files this way if there is no `anaconda-project.yml` file in the same directory.

mcg1969:
@AlbertDeFusco this is passing. We can get this in now if you like, and handle the environment.yml case separately.

AlbertDeFusco:
certainly

Issue #247, opened 2020-01-24
Graceful handling of read-only environments


mcg1969:
If I have a read-only environment in my environment path, and it requires no modifications, anaconda-project should be able to accept it. Unfortunately, it attempts to write to environments for tracking purposes.

mcg1969:
FIxed by #250

PR #250, opened 2020-01-26
Allow environments to be readonly if they have no deviations

mcg1969:
A-P should allow environments to be readonly as long as it does not require any changes to be made. But it expects to be able to write a timestamp file. The authors seemed to have attempted to make it fail silently if the timestamp write fails, but they didn't account for the possibility that creating the directory could fail, too.

We've created a writability test, and mark environments "unfixable" if they have deviations _and_ they are readonly. And we've discarded the requirement that the timestamp write succeeds.


mcg1969:
Fixes #247

PR #251, opened 2020-01-28
Declared jinja2 as a dependency in meta.yml and setup.py

jlstevens:
Title describes the full set of proposed change: `jinja2` is needed for templating introduced in #244. 

In many situations, `jinja2` is an indirect dependency (which is why the Travis and Appveyor tests passed on that PR) but in some situations (e.g `python setup.py develop` in a clean miniconda environment of Windows), `jinja2` will be unavailable. This PR makes `jinja2` an explicit dependency.


jlstevens:
Looks like this PR is ready to merge?

PR #258, opened 2020-03-06
noarch: python

AlbertDeFusco:
The recipe on pkgs/main is noarch. Is there a reason for it *not* to be noarch in the repository as well?


mcg1969:
There is not, as long as it is tested on every platform.

PR #264, opened 2020-04-17
'default' is an alias for the actual default command

AlbertDeFusco:
@bkreider 

This allows us to run a default command (name unkonwn at runtime)
and pass arguments

> anaconda-project run default --anaconda-project-port 8086

The anaconda-project.yml file might look like

```yaml
commands:
  dashboard:
    unix: panel serve main.py
    supports_http_options: true
  api:
    unix: tranquilizer api.py
    supports_http_options: true
```

```
anaconda-project run default # runs panel
anaconda-project run         # runs panel
```

Even though there is not a command called "default" using the
command name default will run the first command, dashboard.

If instead a command is called "default" that will always be
as the default whether the name is included or not (this is
the original behavior and is unchanged)

```yaml
commands:
  dashboard:
    unix: panel serve main.py
    supports_http_options: true
  default:
    unix: tranquilizer api.py
    supports_http_options: true
```

```
anaconda-project run default # runs tranquilizer
anaconda-project run         # runs tranquilizer
```

### DIFF
--- a/anaconda_project/internal/cli/test/test_run.py
+++ b/anaconda_project/internal/cli/test/test_run.py
@@ -493,6 +493,67 @@ def test_run_command_omit_name_use_default(monkeypatch, capsys):
     assert args == ['--version', 'def']
 
 
+def _test_run_nodefault_command(command_line, monkeypatch, capsys, file_assertion=_is_python_exe):
+    executed = {}
+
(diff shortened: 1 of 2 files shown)


bkreider:
Awesome.  That's exactly the behavior I expected!  Looks good to me.

PR #286, opened 2020-09-21
Add command specific variables and define variable priority order

AlbertDeFusco:
There are now 5 places where environment variables can be set for use in `anaconda-project run`. The order listed here is highest priority first, meaning that variables set in any location override the same variable name if exists lower on the list. 

1. Command line invocation
    `MY_VAR='on cli' anaconda-project run`
1. Shell export

    ```
    MY_VAR='exported'
   anaconda-project run
    ```

1. Set within the `anaconda-project.yml` file defined for the command to be run

    ```yaml
    name: project-with-vars
    packages:
      - python=3.7
    commands:
      default:
        variables:
          MY_VAR: 'default command'
        unix: env | grep MY_VAR
    ```

1. Set within the `anaconda-project.yml` at the top level

    ```yaml
    name: project-with-vars
    packages:
      - python=3.7
    variables:
        MY_VAR: 'project'
    commands:
      default:
        unix: env | grep MY_VAR
    ```

1. Within the Conda environment after running `anaconda-project prepare`
    * Using [`conda env config vars`](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables) or through https://github.com/conda/conda/pull/10169 coupled with #275 

## Variable priority examples

The examples here are presented in reverse order from above to demonstrate the override. Each example builds from the previous state.

### Conda env config vars

```yaml
name: vars

packages:
- python=3.7
channels:
- defaults

commands:
  default:
    unix: env | grep MY_VARIABLE
```

```
> anaconda-project prepare
> conda env config vars set MY_VARIABLE='set in conda-meta/state' -p ./envs/default
> anaconda-project run default
MY_VARIABLE=set in conda-meta/state
```

### Top-level project key

Adding `variables` to the project file overrides the Conda env.

```yaml
name: vars

packages:
- python=3.7
channels:
- defaults

variables:
  MY_VARIABLE: 'set in project'

commands:
  default:
    unix: env | grep MY_VARIABLE
```

```
> anaconda-project run default
MY_VARIABLE=set in project
```

### Command-specific variable

```yaml
name: vars

packages:
- python=3.7
channels:
- defaults

variables:
  MY_VARIABLE: 'set in project'

commands:
  default:
    unix: env | grep MY_VARIABLE
  cmd:
    unix: env | grep MY_VARIABLE
    variables:
      MY_VARIABLE: 'set in command: cmd'
```

```
> anaconda-project run default
MY_VARIABLE=set in project
> anaconda-project run cmd
MY_VARIABLE=set in command: cmd
```

### Exported in shell

```
> export MY_VARIABLE='exported'
> anaconda-project run default
MY_VARIABLE=exported
> anaconda-project run cmd
MY_VARIABLE=exported
```

### Set on shell command

```
> MY_VARIABLE='on shell cmd' anaconda-project run default
MY_VARIABLE=on shell cmd
>MY_VARIABLE='on shell cmd' anaconda-project run cmd
MY_VARIABLE=on shell cmd
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,5 @@ htmlcov
 /examples/movies/movies.db
 *.orig
 CHANGELOG.temp
+
+examples/*/envs
--- a/anaconda_project/internal/cli/test/test_prepare.py
+++ b/anaconda_project/internal/cli/test/test_prepare.py
@@ -258,6 +258,45 @@ env_specs:
     assert err == ""
 
 
+def test_prepare_command_default_environment_refresh(capsys, monkeypatch):
+    def mock_conda_create(prefix, pkgs, channels, stdout_callback, stderr_callback):
+        from anaconda_project.internal.makedirs import makedirs_ok_if_exists
+        metadir = os.path.join(prefix, "conda-meta")
+        makedirs_ok_if_exists(metadir)
+        for p in pkgs:
+            pkgmeta = os.path.join(metadir, "%s-0.1-pyNN.json" % p)
+            open(pkgmeta, 'a').close()
+
+    monkeypatch.setattr('anaconda_project.internal.conda_api.create', mock_conda_create)
+
+    def check_prepare_choose_environment(dirname):
+        default_envdir = os.path.join(dirname, "envs", "default")
+        result = _parse_args_and_run_subcommand(
+            ['anaconda-project', 'prepare', '--directory', dirname, '--env-spec', 'default'])
+        assert result == 0
+
+        result = _parse_args_and_run_subcommand(
+            ['anaconda-project', 'prepare', '--directory', dirname, '--env-spec', 'default', '--refresh'])
+        assert result == 0
+
+        assert os.path.isdir(default_envdir)
+
+        foo_package_json = os.path.join(default_envdir, "conda-meta", "nonexistent_foo-0.1-pyNN.json")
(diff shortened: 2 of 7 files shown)


AlbertDeFusco:
I have discovered that Using [Conda `activate.d`](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#saving-environment-variables)  does not work because Anaconda Project does not use the standard Conda activate mechanism

mcg1969:
@AlbertDeFusco should we fix this? That is to say, should we correct how anaconda-project does activation to either use conda's approach directly or reproduce the same steps?

AlbertDeFusco:
Now that `conda run` in version 4.9 supports output streaming [`--no-capture-output`](https://github.com/conda/conda/pull/9646) I'd like to explore wether we could offload the work to `conda run` rather than replicating it.

mcg1969:
That's a good idea, but it raises an important question about how to handle older versions of conda (like those found in AE). As long as we don't break current behavior, but enable newer/better behavior for newer versions of conda, that would be fine.

AlbertDeFusco:
Exactly! It seems like lack of support activate.d has not yet been called out as a bug, so I'm happy to leave this PR as-is and attack activate.d later and only when using against newer conda versions.

PR #292, opened 2020-12-05
Allow multiple environment path directories

mcg1969:
This PR modifies the interpretation of ANACONDA_PROJECT_ENVS_PATH so that multiple environment directories can be considered. Rules:
- The variable is assumed to be a list of paths, separated by `os.pathsep`
- If an entry is empty, it is replaced with `envs`
- Relative paths are considered relative to the project directory (which is why `envs` is the appropriate default)
- If an environment with the given name does not exist anywhere on the path list, it is _created_ in the first directory.
- If the directory is outside of the project directory, it is not removed by cleanup commands.
So for example, suppose you want the default directory to still be the `envs` subdirectory, but you want to allow the child environments in your local Miniconda installation `~/miniconda3` to be considered. Then the value
```
export ANACONDA_PROJECT_ENVS_PATH=:~/miniconda3/envs
```
will produce the desired results.

### DIFF
--- a/anaconda_project/env_spec.py
+++ b/anaconda_project/env_spec.py
@@ -310,7 +310,7 @@ class EnvSpec(object):
     def path(self, project_dir):
         """The filesystem path to the default conda env containing our packages."""
         if self._path is None:
-            for base in os.environ.get('ANACONDA_PROJECT_ENVS_PATH', '').split(':'):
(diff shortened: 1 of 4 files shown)


mcg1969:
The one test failure so far is a timeout on the GA side

AlbertDeFusco:
Very cool.

mcg1969:
Thanks... I'm very close to having persistent sessions work and this is essential for that

PR #302, opened 2020-12-15
Add support for .readonly file

mcg1969:
If a prefix has a `.readonly` file in it, then anaconda-project will assume it is read-only, even if it is not. At bare minimum this is useful for testing


mcg1969:
@AlbertDeFusco I have manually tested this. But I'd like to get this into a tagged build soon, even if before the full readonly support is merged

PR #311, opened 2021-01-15
support default command alias with prepare

AlbertDeFusco:
this enables `anaconda-project prepare --command default` when 'default' is an alias for the first command in the anaconda-project.yml file (see #264 )


mcg1969:
Just curious: if there are three commands, and the _second_ one is called "default", what does `anaconda-project run default` do?

AlbertDeFusco:
It will run the _second_ one. The name `default` takes precedence. If there is no command named `default` the first one in the file is aliased as `default`.

```yaml
commands:
  first:
    unix: echo "I'm first"
  default:
    unix: echo "I'm default"
  last:
    unix: echo "I'm last"
```

```
> anaconda-project run
I'm default
```

```
> anaconda-project run default
I'm default
```

PR #312, opened 2021-02-09
Do not create projectignore if gitignore is present

mcg1969:
Because anaconda-project respects `.gitignore` inside Git repositories there is no need to automatically create `.projectignore` if `.gitignore` is present. It is proving mildly disruptive in AE5 to create this file unnecessarily.

### DIFF
--- a/anaconda_project/archiver.py
+++ b/anaconda_project/archiver.py
@@ -148,6 +148,8 @@ def _load_ignore_file(project_directory, frontend):
 
 def _git_ignored_files(project_directory, frontend):
     if not os.path.exists(os.path.join(project_directory, ".git")):
+        if os.path.exists(os.path.join(project_directory, ".gitignore")):
+            frontend.error("Warning: the .gitignore file is being ignored because this directory is not a Git repository.")
         return []
 
     # It is pretty involved to parse .gitignore correctly. Lots of
--- a/anaconda_project/archiver.py
+++ b/anaconda_project/archiver.py
@@ -218,9 +218,9 @@ def _enumerate_archive_files(project_directory, frontend, requirements):
     if git_filter is None or ignore_file_filter is None:
         return None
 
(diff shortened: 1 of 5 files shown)


anaconda_project/archiver.py
```diff
@@ -148,6 +148,9 @@ def _load_ignore_file(project_directory, frontend):
 
 def _git_ignored_files(project_directory, frontend):
     if not os.path.exists(os.path.join(project_directory, ".git")):
+        if os.path.exists(os.path.join(project_directory, ".gitignore")):
```
mcg1969, line 151: Offers a warning message if this is not a git repo but .gitignore is found. In the future we should parse this file internally instead of relying on git

anaconda_project/archiver.py
```diff
@@ -215,9 +218,9 @@ def _enumerate_archive_files(project_directory, frontend, requirements):
     if git_filter is None or ignore_file_filter is None:
         return None
 
-    plugin_patterns = set()
+    plugin_patterns = {'/anaconda-project-local.yml'}
```
mcg1969, line 221: With this change it is no longer necessary to include this file in projectignore

anaconda_project/projectignore.py
```diff
@@ -11,27 +11,32 @@
...
 /.spyderproject
 """.lstrip()
 
 
 def add_projectignore_if_none(project_directory):
     """Add .projectignore if not found in project directory."""
     filename = os.path.join(project_directory, ".projectignore")
-    if not os.path.exists(filename):
+    gfilename = os.path.join(project_directory, ".gitignore")
+    if not os.path.exists(filename) and not os.path.exists(gfilename):
```
mcg1969, line 39: projectignore is created only if gitignore is not present either

AlbertDeFusco (approved):

PR #331, opened 2021-07-03
progress bar for project download and extract

AlbertDeFusco:
project-download

Give it a try

```
conda install -c defusco/label/dev anaconda-project=0.10.0+4.g35774f5
```

### DIFF
--- a/anaconda_project/archiver.py
+++ b/anaconda_project/archiver.py
@@ -21,8 +21,9 @@ import uuid
 import zipfile
 from io import BytesIO
 from conda_pack._progress import progressbar
+from tqdm import tqdm
 
-from anaconda_project.frontend import _new_error_recorder
+from anaconda_project.frontend import NullFrontend, _new_error_recorder
 from anaconda_project.internal import logged_subprocess
 from anaconda_project.internal.simple_status import SimpleStatus
 from anaconda_project.internal.directory_contains import subdirectory_relative_to_directory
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good! Here is what happened when I tried it:

image

The progress bars worked properly but now I am wondering about those warning messages. I thought it might be due to older dependency versions in my environment but when I created a fresh env I found that they remain. Any ideas?

AlbertDeFusco:
Those warnings are from anaconda-client. Check your `~./continuum/anaconda-client/config.yml` for ssl verify entries. Some of them may be null

jlstevens:
That fixed it, thanks!

I set both `ssl_verify` and `verify_ssl` (how are these things different?) to `true` in `config.yml` and the warnings went away.

Do you have any idea what might have set these flags to false (my only hypothesis is Anaconda Navigator?) as I'm pretty sure I've never manually edited this file before...

AlbertDeFusco:
Something may have changed somewhere recently with anaconda-client. Are you able to test that on a completely fresh Anaconorda/Miniconda install without any of the standard config dirs?

jlstevens:
> Are you able to test that on a completely fresh Anaconorda/Miniconda install without any of the standard config dirs?

I guess I could create a new user on my laptop then install miniconda to see if this reappears...

AlbertDeFusco:
I'll try to test that as well. For now I'm going to let those two failed test combinations slide

Issue #364, opened 2022-02-06
pkg_resources.DistributionNotFound: The 'ruamel_yaml' distribution was not found and is required by anaconda-project


mforbes:
When using an editable install, `pip install -e anaconda-project`, running `anaconda-project` give the error:

```
pkg_resources.DistributionNotFound: The 'ruamel_yaml' distribution was not found and is required by anaconda-project
```

I don't understand all the details -- removing the following from the `bin/anaconda-project` entrypoint fixes it for example:

```
# for compatibility with easy_install; see #2198
__requires__ = 'anaconda-project'
```

however, it is probably related to the name on PyPI [`ruamel.yaml`](https://pypi.org/project/ruamel.yaml/) not `ruamel_yaml`.

Here is how I found this:

```bash
git clone https://github.com/Anaconda-Platform/anaconda-project.git
conda create -n ap3.8 python=3.8 pip
conda activate ap3.8
python -m pip install --no-cache-dir -e anaconda-project
anaconda-project --help
```

<details><summary>Full output</summary>

```bash
(ap3.8) $ anaconda-project --help
Traceback (most recent call last):
  File ".../.conda/envs/ap3.8/bin/anaconda-project", line 33, in <module>
    sys.exit(load_entry_point('anaconda-project', 'console_scripts', 'anaconda-project')())
  File ".../.conda/envs/ap3.8/bin/anaconda-project", line 25, in importlib_load_entry_point
    return next(matches).load()
  File ".../.conda/envs/ap3.8/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File ".../.conda/envs/ap3.8/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 843, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File ".../anaconda-project/anaconda_project/cli.py", line 13, in <module>
    import anaconda_project.internal.cli.main as cli_main
[...]

PR #365, opened 2022-02-06
BUG: Resolves issue #364

mforbes:


AlbertDeFusco:
Oh! I didn't know this variant exists. If you add it to the conda.recipe/meta.yaml and environment.yml files I wonder if the CI tests pass.

https://anaconda.org/main/ruamel.yaml

mforbes:
I added this.  If it works, we might also be able to clean up the import cruft in `yaml_file.py`

mforbes:
Not sure why tests are failing.  I.e. 3.8 on Mac OS X `anaconda_project/internal/test/test_conda_api.py::test_pip_installed` works for me.  I think it might be related to provisioning - the CI is pulling in an old version of `conda` and `pip` for example.

mbargull:
FYI, I've done the same in gh-399 while also removing the then not needed fallback to `ruamel_yaml`.

mcg1969:
Fixed by #413

PR #370, opened 2022-02-16
Fixed CSS link so the theme will look correct.

kathatherine:
Added empty text file to _static folder, which might fix the problem with the custom css not being grabbed by Read the Docs.


kathatherine:
I have no idea what's going on here, but I made the css name more standard anyway, plus the empty text file that all the other _static folders seem to have. Maybe there's actually a good reason.

PR #372, opened 2022-02-16
Update Quick start guide

mattkram:
Updates the quick start guide to walk through running the hvplot interactive demo. Adds a new link to quick start from the documentation homepage.

### DIFF
--- a/docs/source/user-guide/getting-started.rst
+++ b/docs/source/user-guide/getting-started.rst
@@ -7,65 +7,98 @@ Project for the first time.
 
 After completing this guide, you will be able to:
 
-* Create a project containing a Bokeh app.
+* Create a new reproducible project.
 * Run the project with a single command.
 * Package and share the project.
 
 If you have not yet installed and started Project,
 follow the :doc:`Installation instructions <../install>`.
 
-For more information on Bokeh, see `Welcome to Bokeh
-<http://bokeh.pydata.org/en/latest/>`_.
 
+Create a new project
+====================
 
-Creating a project containing a Bokeh app
-===========================================
-
-To create a project called "clustering_app":
+The following steps will create a project called "demo_app":
 
 #. Open a Command Prompt or terminal window.
 
-#. Create a directory called ``clustering_app``, switch to it
-   and then initialize the project::
+#. Initialize the project in a new directory::
 
-     $ mkdir clustering_app
-     $ cd clustering_app
-     $ anaconda-project init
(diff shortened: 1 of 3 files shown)


docs/source/user-guide/getting-started.rst
```diff
@@ -7,65 +7,98 @@ Project for the first time.
...
+.. _hvPlot interactive: https://raw.githubusercontent.com/holoviz/hvplot/master/examples/user_guide/Interactive.ipynb
+.. _quickstart: https://raw.githubusercontent.com/Anaconda-Platform/anaconda-project/master/examples/quickstart/Interactive.ipynb
 
-   A browser window opens, displaying the clustering app.
+Run your project
+================
 
+#. Register a new command to launch the notebook as a `Panel`_ app::
+
+     $ anaconda-project add-command --type unix dashboard "panel serve Interactive.ipynb"
```
AlbertDeFusco, line 93: Perhaps offer the windows variant as well and say `--type unix` is also used on Mac computers.

```
anaconda-project add-command --type windows dashboard "panel serve Interactive.ipynb"
```

AlbertDeFusco:
love this!

docs/source/user-guide/getting-started.rst
jbednar: Here's a good example of where we'd like not to have to specify the platform! 

Also, without https-options, I think this will only work locally and not inside a container...

mattkram: Good call on platforms. I'll add a note for windows users and totally agree. I was surprised to be asked about what type of dashboard when I didn't include `--type unix`. That's the type of implicit behavior we should remove.

Regarding Docker, I sort of expected this would only work for local. I can put a comment about that. Would personally prefer just a note instead of introducing more concepts in quickstart.

AlbertDeFusco: Yes, dockerize would require `supports_http_options`

docs/source/user-guide/getting-started.rst
```diff
@@ -7,65 +7,98 @@ Project for the first time.
...
+
+.. _Panel: https://panel.holoviz.org
+
+#. Run your new project::
+
+     $ anaconda-project run
```
AlbertDeFusco, line 99: If you recommend this approach the browser will open automatically

```
anaconda-project run dashboard --show
```

mattkram: Thank you! I was looking for that option.

Issue #387, opened 2022-08-01
[ENH] Implicit "defaults" channel


AlbertDeFusco:
Starting in version 0.11.0 condarc "channels:" list is ignored. 

* If a project yaml does not have any "channels:" listed it is now unusable.
* Some project yaml files have "channels:" but would implicitly rely on "defaults"
    * in an unspecified/tested/studied order

Proposal:
* always use "defaults" with and without explicit channel lists
* implement Conda env create's "nodefaults": https://github.dev/conda/conda/blob/9bf4c1721da4d44739546109a78fb2edb40af95b/conda_env/installers/conda.py#L19
* if "defaults" is not listed in the ordered-list of channels it will go **????**

Goal:
* ensure some backward compatibility while still not relying on condarc state

AlbertDeFusco:
@jbednar  and @jlstevens where should `defaults` go when not supplied in `channels` and when `nodefaults` is not also specified? First or last?

jbednar:
I vote for implementing`nodefaults` and for otherwise assuming `defaults` is in the list.

> @jbednar and @jlstevens where should defaults go when not supplied in channels and when nodefaults is not also specified? First or last?

Wherever anaconda-project  behaves as it did before. Did it previously act like it was implicitly first, or implicitly last?

jlstevens:
I will confirm it for myself but I believe it was implicitly last (as Albert mentioned above).

AlbertDeFusco:
merged and released in 0.11.1

PR #388, opened 2022-08-02
implicit use of "defaults" channel

AlbertDeFusco:
This addresses #387 

The "defaults" is always appended to channels: list. This is disabled by adding a dummy channel called nodefaults.

@jbednar and @jlstevens I have dev build of the implementation of nodefaults.

```
conda install -c defusco/label/dev anaconda-project=0.11.0+2
```

I am working through the current test suite and will amend it for this case as well. Here is my test:

In version 0.11.0 the following project will not prepare or lock

```yaml
name: nodefaults

channels:
  - conda-forge

packages:
  - python=3.8
  - conda-token
  - ensureconda
```

the error from 0.11.0 is 

```
The following packages are not available from current channels:

  - conda-token

Current channels:

  - https://conda.anaconda.org/conda-forge/osx-arm64
  - https://conda.anaconda.org/conda-forge/noarch
```

Note that Conda's use of `nodefaults` is not identical to what is done above. In Conda

1. `nodefaults` can only be used within an environment.yml file
2. instead of appending just the `defaults` channel all channels from users .condarc file are added

Here ONLY `defaults` is added, any channels provided in condarc are still ignored.

With this dev release the above YAML file will lock and prepare.

To recover the behavior of 0.11.0 add a channel called `nodefaults`. This will prevent `defaults` from being added at the end of your channels list

```yaml
name: nodefaults

channels:
  - conda-forge
  - nodefaults

packages:
  - python=3.8
  - conda-token
  - ensureconda
```

### DIFF
--- a/docs/source/user-guide/tasks/work-with-packages.rst
+++ b/docs/source/user-guide/tasks/work-with-packages.rst
@@ -114,7 +114,7 @@ Package Channels
   or in the project YAML file.  To support reproducible projects that build the same way for different users, Anaconda Project will not respect channels declared in your ``.condarc`` file.
 
 .. note::
-  *Backwards compatibility fix in version 0.11.1*. The ``defaults`` channel is always included when packages are installed
+  *Backwards compatibility fix in version 0.11.1*. The ``defaults`` channel is always appended when packages are installed
   or locked even if it is not specified in the ``channels:`` list. To avoid searching over the ``defaults`` channel add the chanel ``nodefaults``.
 
 
--- a/docs/source/user-guide/tasks/work-with-packages.rst
+++ b/docs/source/user-guide/tasks/work-with-packages.rst
@@ -114,14 +114,15 @@ Package Channels
   or in the project YAML file.  To support reproducible projects that build the same way for different users, Anaconda Project will not respect channels declared in your ``.condarc`` file.
 
 .. note::
-  *Backwards compatibility fix in version 0.11.1*. The ``defaults`` channel is always appended when packages are installed
-  or locked even if it is not specified in the ``channels:`` list. To avoid searching over the ``defaults`` channel add the chanel ``nodefaults``.
+  *Backwards compatibility fix in version 0.11.1*. The ``defaults`` channel is always appended to the list of channels
+  even if it is not specified in the ``channels:`` list in the project file or with the ``-c`` flag using the CLI.
+  If the ``defaults`` channel is specified no change is made. To avoid including the ``defaults`` channel add the
+  channel name ``nodefaults``.
 
 
 Up till now we have not instructed Conda to install packages from specific channels, so all packages are installed from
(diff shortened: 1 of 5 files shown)


jlstevens:
Looks like backwards compatibility is (partially) restored! https://github.com/holoviz/holoviz/pull/314

I say 'partially' as we have been using `nodefaults` in project yamls even though it was having no effect. Clearly we (mistakenly) *thought* it was doing something (which is now true with this PR). 

In this instance, I am happy to chalk it up to user error and this is *not* something we should try to maintain backwards compatibility for - having the implicit `defaults` is the more important fix for backwards compatibility imho than restoring the old behavior of ignoring `nodefaults`. Do you agree @jbednar ?

jbednar:
Yep! This sounds perfect! Just need to merge and make a new release ASAP so our projects start working again!

AlbertDeFusco:
Ok. I'll fix the tests and update docs

docs/source/user-guide/tasks/work-with-packages.rst
```diff
@@ -113,6 +113,11 @@ Package Channels
   *Breaking Change in version 0.11.0*. All channels you wish to search through for packages must be supplied on the CLI
   or in the project YAML file.  To support reproducible projects that build the same way for different users, Anaconda Project will not respect channels declared in your ``.condarc`` file.
 
+.. note::
+  *Backwards compatibility fix in version 0.11.1*. The ``defaults`` channel is always included when packages are installed
```
maximlt, line 117: To be more explicit?

```suggestion
  *Backwards compatibility fix in version 0.11.1*. The ``defaults`` channel is always appended when packages are installed
```

maximlt (commented):
As I understand it from the discussion before 0.11 the channels from condaRC were appended to the project channels, and I understand this PR resets that behavior. This sounds fine to me. If it's not yet documented, I believe it would be good to mention that somewhere.

Also, is the implicit `defaults` channel appended even if `defaults` is listed in the project channels? And if so, which one takes precedence?

AlbertDeFusco:
This will only add defaults, not all channels listed in condarc.

Also, I will fix this to not adjust channels if defaults is already present

AlbertDeFusco:
I've made the change to avoid adding `defaults` if already present so that it won't change your priorities.

If you like please try the dev release with the latest commits

```
conda install -c defusco/label/dev anaconda-project=0.11.0+11
```

docs/source/user-guide/tasks/work-with-packages.rst
jlstevens: Given the latest discussion, it sounds like there should also be a sentence along the lines of:

'If the defaults channel is explicitly specified in the channel list, the stated channel ordering will be used and the defaults channel will not be appended again at the end of the list.'

PR #389, opened 2022-10-16
Changed .format to f-string for better readability

AdityaPunetha:
Changed .format to f-string for better readability and performance as per PEP 498


mcg1969:
At this late state of this library's life we should avoid unnecessary changes

PR #401, opened 2024-02-09
Display download bar in all cases and make the timeout configurable via an env var

maximlt:
We have observed two issues when files are downloaded in the *prepare* phase:

1. Responses without `Content-Length` lead to no download bar being displayed at all
2. Downloads longer than 10 minutes time out

This PR addresses 1) by displaying a progress bar in any case, so the user is aware of some sort of progress. They can also see if it hands and decide to abort if so.

image

2 isn't fully addressed, instead, the default timeout of 600s is made configurable via the `ANACONDA_PROJECT_DOWNLOADS_TIMEOUT` environment variable. I guess the best approach would be to implement a timeout that is raised only when nothing happens between receiving two chunks, instead of setting a global request timeout.

### DIFF
--- a/anaconda_project/internal/http_client.py
+++ b/anaconda_project/internal/http_client.py
@@ -31,6 +31,7 @@ class FileDownloader(object):
         self._client = None
         self._errors = []
         self._progress = None
+        self._progress_kwargs = None
 
     @gen.coroutine
     def run(self):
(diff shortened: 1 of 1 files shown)


maximlt:
@AlbertDeFusco I've been told to ping you! :)

PR #404, opened 2024-06-14
Remove documentation referring to Anaconda Cloud 

bkentropy:
Remove documentation referring to Anaconda Cloud (anaconda.org) because it will no longer supported. I'd like to remove the code too. But there's two good reasons not to quite yet.
1. We have some further dispensing of information alerting users that `anaconda-project` will no longer be able to upload or download to anaconda.org.
2. The tests and deployment in the repo needs some love. The package can't build in CI, and locally I had to pin versions of `python` and `anaconda-client` to get tests to pass. Hopefully we can address this soon.


docs/source/getting-started.rst
```diff
@@ -100,36 +100,6 @@ Run your project
...
-To share this project with a colleague:
-
-#. Archive the project::
-
-     $ anaconda-project archive interactive.zip
-
-#. Send the archive file to your colleague.
-
-You can also share a project by uploading it to Anaconda Cloud.
-For more information, see :doc:`user-guide/tasks/share-project`.
```
AlbertDeFusco, line 115: Actually, this is the only portion that needs to be removed. Creating a zip and sharing by means other than anaconda.org should still be documented.

AlbertDeFusco (dismissed):

docs/source/getting-started.rst
bkentropy: Good point, will fix 👍

mcg1969 (approved):

PR #427, opened 2026-05-13
fix(export-pixi): translate env vars and unify unix/windows commands

mcg1969:
## Summary

* `pixi run <task>` was failing on exported `pixi.toml` files because anaconda-project command strings carry references like `${PROJECT_DIR}`, `%PROJECT_DIR%`, and explicit `${CONDA_PREFIX}/bin/...` paths that pixi can't resolve.
* This PR rewrites command strings during export so they execute correctly under pixi's deno_task_shell on every platform.

## What changed

Three normalization passes run on every command line emitted to the `[tasks]` / `[feature.*.tasks.*]` sections:

* **Env-var translation** (`_translate_command_env_vars`): `${VAR}`/`$VAR`/`%VAR%` references are normalized to the deno_task_shell `${VAR}` form. Well-known anaconda-project vars are mapped to their pixi equivalents:
    * `PROJECT_DIR` → `PIXI_PROJECT_ROOT`
    * `CONDA_ENV_PATH` → `CONDA_PREFIX`
    * `CONDA_PREFIX`, `CONDA_DEFAULT_ENV`, `PATH` pass through (pixi sets all three).
    * Vars declared by the project itself (`variables:`, `downloads:`, `services:`) pass through unchanged.
    * Unknown references are flagged in a `# unresolved env var(s): ...` comment.
* **Windows command normalization** (`_windows_to_deno_shell`): backslashes inside path-shaped tokens are swapped to forward slashes. When both `unix:` and `windows:` variants exist, they're compared after translation; matching forms collapse to a single task, and any remaining divergence is flagged in a comment. Windows-only commands now translate and emit instead of being dropped or emitted as raw cmd.exe syntax.
* **CONDA_PREFIX path stripping** (`_strip_conda_prefix_paths`): `${CONDA_PREFIX}/bin/X`, `/Scripts/X`, `/Library/bin/X`, etc. reduce to bare `X`, since pixi's task activation already prepends those directories to PATH. This makes the cross-platform forms agree far more often — e.g. `${CONDA_PREFIX}/bin/python` and `%CONDA_PREFIX%\python.exe` both become plain `python`.

### Example

The `Hello Anaconda Enterprise` template exports correctly now:

```toml
[feature.python-37.tasks.default]
cmd = "python ${PIXI_PROJECT_ROOT}/hello.py"
```

Previously the cmd was `python ${PROJECT_DIR}/hello.py`, which deno_task_shell does not expand.

## Test plan

* [x] 24 new unit tests covering env-var translation, windows normalization, conda-prefix stripping, unix/windows unification, and unknown-var flagging
* [x] All 42 tests in `test_pixi_export.py` pass
* [x] Hello Anaconda Enterprise template (the original failing case) now renders with `${PIXI_PROJECT_ROOT}` and runs under `pixi run <task>`
* [ ] Test against a project with diverging unix/windows commands and confirm the divergence comment is informative
* [ ] Test against a project that uses `${CONDA_PREFIX}/bin/...` explicitly and confirm the stripped form runs

### DIFF
--- a/anaconda_project/internal/pixi_export.py
+++ b/anaconda_project/internal/pixi_export.py
@@ -102,21 +102,192 @@ def _write_dependencies(lines, conda_packages, pip_packages, indent=''):
         lines.append('')
 
 
-def _command_to_task(command):
+# anaconda-project sets these in the runtime environment for every task; pixi
+# provides equivalents that we can substitute in command strings so that
+# `pixi run <task>` resolves them the same way `anaconda-project run` does.
+# Vars that pixi sets natively (CONDA_PREFIX, PATH, CONDA_DEFAULT_ENV) map to
+# themselves and just need to survive the syntax conversion below.
+_ANACONDA_PROJECT_ENV_VAR_MAP = {
+    'PROJECT_DIR': 'PIXI_PROJECT_ROOT',
+    'CONDA_ENV_PATH': 'CONDA_PREFIX',
+    'CONDA_PREFIX': 'CONDA_PREFIX',
+    'CONDA_DEFAULT_ENV': 'CONDA_DEFAULT_ENV',
+    'PATH': 'PATH',
+}
+
+# Matches ${VAR}, $VAR, and %VAR% references in a command string. The Windows
+# %VAR% form is included because anaconda-project commands routinely carry
+# both unix and windows command lines, and we may emit either.
+_ENV_VAR_REF_RE = re.compile(
+    r'\$\{(?P<braced>[A-Za-z_][A-Za-z0-9_]*)\}'
+    r'|\$(?P<bare>[A-Za-z_][A-Za-z0-9_]*)'
+    r'|%(?P<windows>[A-Za-z_][A-Za-z0-9_]*)%'
+)
+
+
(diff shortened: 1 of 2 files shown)


A-Aron-T (approved):
Tested locally against several project configurations. The env-var translation produces correct pixi-runnable commands in all cases I tried:

- `${PROJECT_DIR}` → `$PIXI_PROJECT_ROOT` ✓
- `${CONDA_PREFIX}/bin/python` → bare `python` ✓
- Unix + Windows forms unify cleanly when equivalent after normalization ✓
- Declared project variables pass through without false "unresolved" warnings ✓
- Unknown vars get flagged with actionable comment ✓
- Windows-only commands translate to deno_task_shell form instead of being emitted raw ✓

## Notes

1. The ambiguous-suffix case (`${PIXI_PROJECT_ROOT}suffix`) still emits braces which deno_task_shell rejects — acknowledged in the code as unsolvable, and the original command had the same issue. Fine.

2. This is a significant quality improvement for converted projects. Most real-world anaconda-project commands use `${PROJECT_DIR}` and previously produced broken pixi tasks.

LGTM.

PR #428, opened 2026-05-13
feat(export-pixi): conversion conventions for env layout and prepare tasks

mcg1969:
## Summary

Several iterations on the `export-pixi` command, driven by real-world testing against converted projects (Glaciers, NYC Taxi). The PR started as a single-bug fix for env_spec naming but grew into a coherent set of conventions for what a converted pixi.toml should look like and how the runtime behavior should mirror `anaconda-project`.

### Environment layout

* **Single non-default env_spec** (e.g. `sampleproj:`): packages live in top-level `[dependencies]` (the default feature) and the named env inherits via `features = ["sampleproj"]`. Pixi's mandatory implicit `default` env carries the same packages, so `pixi install` works without `-e`.
* **`default` mixed into multi-env**: comment-out its slot in `[environments]` (`# default  (pixi creates this implicitly...)`) and fold its packages into top-level `[dependencies]`. Avoids redeclaring an env pixi auto-creates anyway.
* **env_specs emit in source order**, not alphabetical. The first uncommented entry in `[environments]` is the user's intended default — extractable with one line of `awk` for downstream tooling.
* **`solve-group` removed**. anaconda-project doesn't assume environments solve together; the converted manifest shouldn't either.

### `prepare` task convention

* Every converted project gets a `prepare` task, scoped to the default env_spec's feature (or top-level when the manifest has no `[environments]` table). The task does double duty: it carries the download logic when there are downloads, and it serves as a converted-project marker plus env-selection entry point in every case.
* The default env is whatever `project.default_env_spec_name` resolves to — typically the first declared env_spec. A literal `default` env_spec wins, since pixi's implicit `default` env materializes from the default feature anyway.
* Other env_specs do **not** get a prepare task. anaconda-project's global `downloads:` apply to every env, but we only need to fetch them once. Keeping `prepare` scoped to a single feature also sidesteps pixi's "task is defined in multiple environments" prompt entirely.
* When the source declares `downloads:`, prepare invokes a sibling helper `ap_download.py` (one line per download) instead of inlining urllib. Helper is written next to `pixi.toml` at conversion time and uses fixed positional args: `<url> <filename> [<description>]`.
* When there are no downloads, prepare is a no-op `echo` — small footprint, but its presence is what makes `pixi run prepare` resolve to a non-default-named env (e.g. `sampleproj`) automatically.
* Description field omitted from prepare tasks — pixi appends it to the banner as clutter. Run with `pixi run -q prepare` for clean output.

### `python` for downloads

* Helper invoked as `python3 ap_download.py ...` so it works with system `python3` when the env doesn't declare its own python.
* If a download-needing env lacks `python`, a `# WARNING:` block at the top of `pixi.toml` lists the affected envs. Same warning is also printed to stderr at conversion time.
* No silent mutation of the user's dep list.

## Files

* `anaconda_project/internal/pixi_export.py` — exporter logic.
* `anaconda_project/internal/ap_download.py` — sibling helper script.
* `anaconda_project/project_ops.py` — copies the helper next to pixi.toml; surfaces warnings.

## Test plan

* [x] 52 unit tests in `test_pixi_export.py` pass.
* [x] Manual verification against Glaciers (`sampleproj` env, no downloads).
* [x] Manual verification against NYC Taxi (`sampleproj` env, one download with description).
* [ ] `pixi install -e sampleproj` builds the env.
* [ ] `pixi run prepare` fetches the download.
* [ ] `pixi run dashboard` (or equivalent named task) auto-resolves to the named env without `-e`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,11 @@ Thumbs.db
 # Project-specific
 anaconda-project-local.yml
 anaconda-project-prs/
+
+# Local AI tooling settings
+.claude/
+
+# Test artifacts produced by the local test runner
+cov.xml
+fake_project/
+fake_project.zip
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,20 @@
+# Read the Docs configuration. See:
+# https://docs.readthedocs.com/platform/stable/config-file/v2.html
+#
+# Read the Docs requires this file (in the repository root) to build
+# the docs. Sphinx config and dependencies live under docs/.
+
+version: 2
+
+build:
+  os: ubuntu-24.04
+  tools:
+    # miniforge3 supplies conda + the conda-forge channel by default,
+    # which matches the resolution we get locally with the env file.
+    python: "miniforge3-latest"
+
+conda:
+  environment: docs/environment.yml
+
+sphinx:
+  configuration: docs/source/conf.py
--- a/anaconda_project/archiver.py
+++ b/anaconda_project/archiver.py
@@ -20,8 +20,12 @@ import tempfile
 import uuid
 import zipfile
 from io import BytesIO
-from conda_pack._progress import progressbar
-from tqdm import tqdm
+
+# `conda_pack` and `tqdm` are deferred to the call sites below
+# (`_archive_project_dir`, `_extract_files`). Importing them eagerly
+# adds tens of milliseconds to every `import anaconda_project.project`,
+# which `publication_info` and other lightweight consumers pay for no
+# reason — neither package is touched on the read path.
 
 from anaconda_project.frontend import NullFrontend, _new_error_recorder
 from anaconda_project.internal import logged_subprocess
@@ -259,6 +263,7 @@ def _leaf_infos(infos):
 
 
 def _write_tar(archive_root_name, infos, filename, compression, packed_envs, frontend):
+    from conda_pack._progress import progressbar
     if compression is None:
         compression = ""
     else:
@@ -294,6 +299,7 @@ def _write_tar(archive_root_name, infos, filename, compression, packed_envs, fro
 
 
 def _write_zip(archive_root_name, infos, filename, packed_envs, frontend):
+    from conda_pack._progress import progressbar
     with zipfile.ZipFile(filename, 'w') as zf:
         for info in _leaf_infos(infos):
             arcname = os.path.join(archive_root_name, info.relative_path)
@@ -446,6 +452,7 @@ def _extractall_chmod(zf, destination):
 
 
 def _extract_files_zip(zip_path, src_and_dest, frontend):
+    from tqdm import tqdm
     # the zipfile API has no way to extract to a filename of
     # our choice, so we have to unpack to a temporary location,
     # then copy those files over.
@@ -472,6 +479,7 @@ def _extract_files_zip(zip_path, src_and_dest, frontend):
 
 
 def _extract_files_tar(tar_path, src_and_dest, frontend):
+    from tqdm import tqdm
     with tarfile.open(tar_path, mode='r') as tf:
         if isinstance(frontend.underlying, NullFrontend):
             src_and_dest = tqdm(src_and_dest, desc='Extract ')
--- a/anaconda_project/internal/ap_download.py
+++ b/anaconda_project/internal/ap_download.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""Helper invoked by `prepare` tasks in pixi.toml files converted from
+anaconda-project.yml. Fetches one download at a time, skipping if the
+target file already exists.
+
+Usage:
+    python ap_download.py <url> <filename> [<description>]
+
+Pure stdlib: urllib + os, so it runs in any conda env that has python.
+"""
+from __future__ import absolute_import, print_function
+
+import os
+import sys
+import urllib.request
+
+
+def main(argv):
+    if len(argv) not in (2, 3):
+        print('usage: python ap_download.py <url> <filename> [<description>]',
+              file=sys.stderr)
+        return 2
+
+    url = argv[0]
+    filename = argv[1]
+    description = argv[2] if len(argv) == 3 else filename
+
+    if os.path.exists(filename):
+        print('[prepare] {}: exists at {}'.format(description, filename))
+        return 0
+
+    parent = os.path.dirname(filename)
+    if parent:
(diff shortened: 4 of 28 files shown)


A-Aron-T:
Tested locally against the rollup branch (all PRs combined). Exported several sample projects including Panel dashboards, NLP APIs, and a project with downloads.

**Env layout conventions are solid** — single non-default env_spec correctly promotes packages to top-level `[dependencies]` with the feature scoping for tasks. The `prepare` task marker works as expected for env-selection entry point (forces `pixi run prepare` to resolve to the correct env when the name isn't "default").

**`ap_download.py` helper verified** — exported a project with two `downloads:` entries. The `prepare` task correctly calls `python3 ap_download.py <url> <filename> [<description>]` for each download. The generated helper is pure stdlib (urllib + os), idempotent (skips existing files), creates parent directories, and weighs in at ~1KB. Good design — no heavy deps needed on the download path.

96 pixi export tests pass, plus the downstream `publication_info()` round-trip (export → read back) returns correct env_spec names and command mappings.

Looks good from our side.

mcg1969:
Thanks for the local verification — good to know the `ap_download.py` helper holds up against real `downloads:` projects and that the env-layout heuristics line up with what your tooling expects. No changes from this comment; flagging it as acknowledged.

mcg1969:
**Folded in: #429** (`feature/pixi-channel-defaults` → `feature/pixi-single-named-env`)

This PR is becoming a rollup. Tracking what's been folded:

- [x] #429 — channel defaults expansion + project_info notebook/env_specs fixes (3 commits)
- [ ] #430 — http options + publication_info improvements
- [ ] #435 — `--use-default` and `--add-current-platform` flags
- [ ] #431 — docs: project status, pixi support page, modern feedback channels

Awaiting CI before folding the next one in.

mcg1969:
**Folded in: #430** (`feature/pixi-http-options` → `feature/pixi-single-named-env`)

Rollup checklist:

- [x] #429 — channel defaults expansion + project_info notebook/env_specs fixes
- [x] #430 — http options + publication_info improvements (13 commits, including the `info` subcommand and `test_main_help` fix)
- [ ] #435 — `--use-default` and `--add-current-platform` flags
- [ ] #431 — docs: project status, pixi support page, modern feedback channels

Awaiting CI before folding the next one in.

mcg1969:
**Folded in: #435** (`feature/pixi-use-default` → `feature/pixi-single-named-env`)

Rollup checklist:

- [x] #429 — channel defaults expansion + project_info notebook/env_specs fixes
- [x] #430 — http options + publication_info improvements
- [x] #435 — `--use-default` and `--add-current-platform` flags (4 commits, plus public API surface for downstream consumers)
- [ ] #431 — docs: project status, pixi support page, modern feedback channels

Awaiting CI before folding the final one in.

mcg1969:
**Folded in: #431** (`feature/docs-update` → `feature/pixi-single-named-env`)

Rollup checklist — complete:

- [x] #429 — channel defaults expansion + project_info notebook/env_specs fixes
- [x] #430 — http options + publication_info improvements
- [x] #435 — `--use-default` and `--add-current-platform` flags
- [x] #431 — docs: project status, pixi support page, modern feedback channels (10 commits, including `.readthedocs.yaml`, the docs build env, the pixi support page, and the install/copyright refresh)

This PR is now the consolidated rollup. Awaiting final CI, then will request @A-Aron-T's approval for the master merge.

A-Aron-T (approved):
Tested locally — all 188 tests pass (96 export, 75 project_info, 10 info CLI, 7 main CLI). End-to-end export verified with single-named-env project containing HTTP options and downloads.

## Findings

### `[tool.anaconda.commands]` not emitted (follow-up needed)

The exporter doesn't generate `[tool.anaconda.commands]` in the output pixi.toml. `project_info.py` reads this section for explicit `supports_http_options` metadata; without it, heuristic detection (`_looks_like_http`) kicks in. The heuristic catches panel/bokeh/jupyter/flask/uvicorn/streamlit but misses generic `python app.py` commands that had `supports_http_options: true` in the original manifest.

Verified locally: exported a project with `python app.py` + `supports_http_options: true`, read it back via `anaconda-project info --json` — pixi path reports `supports_http_options: false` for that command while the anaconda-project path correctly reports `true`.

**Not a merge blocker** — the `args` field in publication_info is what `start_http_server.py` actually uses for positional arg injection at deploy time, and that works correctly regardless of `supports_http_options`. The flag primarily affects UI deploy dialogs (which commands appear as deployable services). Fix is additive: emit `[tool.anaconda.commands.X]` with `supports_http_options = true` alongside the existing task definition.

### Minor notes (non-blocking)

- `ap_download.py` uses deprecated `urllib.request.urlretrieve` — works fine but may need updating if removed in future Python
- No checksum verification for downloads (original manifest supports md5/sha256)
- `_format_dep` for inline-table pixi specs (`{version = ">=2.0"}`) returns just the package name with no version constraint — correct behavior per the code's intent but worth knowing for downstream consumers comparing expected vs installed

mcg1969:
Thanks for the detailed review. Pushed [a0e1503](https://github.com/anaconda/anaconda-project/commit/a0e1503) addressing two of the three findings:

**`_format_dep` inline-table specs** — Fixed. Pixi `numpy = { version = ">=2.0" }` now reads back as `"numpy>=2.0"` instead of dropping to bare `"numpy"`. Other dict keys (channel, build) don't have a place in the conda spec string form, so they're ignored. New tests cover the dict + operator, dict + bare-version, dict + wildcard, and dict-without-version cases.

**Download checksum verification** — Fixed. The exporter now threads `hash_algorithm` / `hash_value` from each `DownloadRequirement` through to the `ap_download.py` helper as `--<algo> <hex>` flags. The helper verifies after download, removes the bad file on mismatch (since it came from the network), and refuses without removing when an existing on-disk file fails verification (since the user may have edited it intentionally). Bare-url downloads with no checksum keep the old 3-positional invocation; the helper CLI is back-compatible. Added 17 unit tests for the helper covering arg parsing, both verification paths, and the skip-existing case.

**`[tool.anaconda.commands]` / `supports_http_options` round-trip** — Discussed in Slack. Short version: `supports_http_options` is a *pre-export* contract about whether the underlying app accepts the `--anaconda-project-X` flag namespace, not a deployability signal. Once exported, that contract is already baked into the rendered task body (the gated flags), so there's no actionable round-trip use for the bit post-export. Skipping for now; happy to revisit if a concrete consumer needs it.

CI running. The other minor `urlretrieve`-deprecation note is left as-is — pending-deprecation since 3.6, no removal date, still valid in 3.14.

PR #429, opened 2026-05-13
feat(export-pixi): channel defaults; project_info notebook + env_specs fixes

mcg1969:
## Summary

Stacked on #428. Three commits:

1. **Channel defaults expansion** (described below).
2. **Notebook-detection narrowing in project_info**: `_infer_notebook` previously flagged any command with a `.ipynb` token as a notebook, which mis-classified `panel serve foo.ipynb` and `voila foo.ipynb` as notebooks. Now only commands that explicitly launch Jupyter (`jupyter notebook|lab`) are classified that way — matching what anaconda-project's `notebook:` shorthand expands to. Other tools that consume an `.ipynb` are still picked up as HTTP services.
3. **Two env_specs fixes for pixi projects** in `publication_info`:
   * Envs declared with `no-default-feature = true` previously inherited the top-level `[dependencies]`. They now correctly start from an empty package list and only collect packages from the features they list. Surfaced by `workbench-sample-projects/projects/attractors/pixi.toml`, which has `testenv = { features = ["testenv"], no-default-feature = true }`.
   * Pixi always materializes a `default` env even when `[environments]` declares only other envs. `publication_info` now surfaces `default` unconditionally and honors a user-declared `default = { features = [...], ... }` instead of silently overwriting it with the top-level dependencies.

## Channel defaults expansion

* Pixi has no `defaults` meta-channel and no equivalent of conda's `default_channels` config.
* The previous exporter either passed `defaults` through verbatim (later breaking at install time) or hard-coded `"conda-forge"` when the yml declared no channels at all, ignoring the user's actual conda configuration.
* This PR resolves `defaults` by shelling out to `conda config --show default_channels` and substituting the result.

## Behavior

* **No channels in the yml** → populate from `conda config --show default_channels` (typically `https://repo.anaconda.com/pkgs/main` and `.../pkgs/r`, but enterprise users with custom `.condarc` get their own mirror URLs).
* **`defaults` mixed with other channels** → expand only that entry; preserve the rest in source order. Dedup so an explicit channel isn't re-added by the expansion.
* **No `defaults` reference at all** → no conda invocation; channels pass through.

## Why not `--json`?

`conda config --show --json` renders URL fields as nested urlparse dicts (`{"scheme": "https", "netloc": "...", ...}`) instead of the original strings. The plain text output is `key:` followed by `  - <value>` list entries — easy to parse and stable.

## Hard-error path

If `conda` isn't on PATH or fails to invoke, the exporter raises `CondaNotAvailableError` and `project_ops.export_pixi` returns a failure status. Nothing is written. Faithful conversion requires resolving `defaults`, so we don't fall back silently.

## Test plan

* [x] Unit tests for `_expand_defaults_in_channels` (no-defaults, in-place, mixed, dedup, repeat).
* [x] End-to-end test verifying `defaults` is expanded in the converted manifest.
* [x] End-to-end test verifying the no-channels case populates from conda's defaults.
* [x] Conda-missing test asserts `CondaNotAvailableError` is raised.
* [x] Notebook-classification tests cover the `panel serve foo.ipynb` / `voila foo.ipynb` cases.
* [x] `no-default-feature = true` test verifies the env's package list excludes top-level deps but includes the listed feature's deps.
* [x] `default`-always-present test verifies `default` is in `env_specs` even when `[environments]` only lists other envs, and honors a user-declared `default`.
* [x] All `test_project_info.py` and `test_pixi_export.py` tests pass.
* [ ] Verify the converted Glaciers / NYC Taxi pixi.toml installs cleanly when `defaults` is expanded.

### DIFF
--- a/anaconda_project/internal/pixi_export.py
+++ b/anaconda_project/internal/pixi_export.py
@@ -10,11 +10,82 @@ from __future__ import absolute_import, print_function
 
 import os
 import re
+import shutil
+import subprocess
 
 from anaconda_project.requirements_registry.requirement import EnvVarRequirement
 from anaconda_project.requirements_registry.requirements.conda_env import CondaEnvRequirement
 from anaconda_project.requirements_registry.requirements.download import DownloadRequirement
 from anaconda_project.requirements_registry.requirements.service import ServiceRequirement
+
+
+class CondaNotAvailableError(Exception):
+    """Raised when the exporter needs `conda` for channel resolution but
+    can't find or invoke it. Wraps the underlying reason so the CLI can
+    surface a useful failure status."""
+
+
+def _resolve_default_channels():
+    """Return the list of URLs that `defaults` expands to, taken from the
+    user's local ``conda config --show default_channels``.
+
+    Pixi has no notion of a `defaults` meta-channel, so we have to rewrite
+    every occurrence to the URLs it would resolve to under conda. We avoid
+    `--json` because conda renders URL fields as nested urlparse dicts in
+    that mode; the plain YAML-ish output is easier to parse and is what
+    `conda config --show` emits by default.
+    """
+    if not shutil.which('conda'):
+        raise CondaNotAvailableError(
+            "`conda` not found on PATH; required to expand the `defaults` "
+            "channel into concrete URLs.")
+    try:
+        proc = subprocess.run(
+            ['conda', 'config', '--show', 'default_channels'],
+            check=True, capture_output=True, text=True)
+    except (OSError, subprocess.CalledProcessError) as e:
+        raise CondaNotAvailableError(
+            "Failed to invoke `conda config --show default_channels`: {}".format(e))
+
+    urls = []
+    in_block = False
+    for line in proc.stdout.splitlines():
+        if line.startswith('default_channels:'):
+            in_block = True
+            continue
+        if in_block:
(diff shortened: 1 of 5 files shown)


A-Aron-T:
Tested locally. Channel defaults expansion works correctly — `defaults` resolves to `https://repo.anaconda.com/pkgs/main` + `https://repo.anaconda.com/pkgs/r` via `conda config --show default_channels`. Verified this produces valid pixi channel URLs.

The notebook-detection narrowing is a good fix — `panel serve foo.ipynb` was previously mis-classified as a notebook command by `_infer_notebook`. Now only explicit `jupyter notebook|lab` launches get the `notebook` tag in `publication_info()`. Confirmed this in our Panel dashboard test project (tagged `[http]` not `[notebook, http]`).

Also confirmed `no-default-feature = true` env_specs don't inherit top-level `[dependencies]` — 75 project_info tests pass covering that case plus the "default env always materialized" behavior.

96 pixi export tests + 75 project_info tests pass.

Looks good.

mcg1969:
Thanks. Glad the `panel serve foo.ipynb` re-tagging behavior matches what you needed — that case was the main motivator for narrowing `_infer_notebook` to explicit `jupyter notebook|lab` launchers. No changes needed; acknowledged.

PR #430, opened 2026-05-14
feat(export-pixi): http options + publication_info improvements

mcg1969:
## Summary

Stacked on #429. HTTP-options support across all three command shapes, several `publication_info` fixes for downstream consumers, and a perf fix that cuts cold-start cost on the read path by ~80%.

## HTTP options translation in `pixi_export`

`supports_http_options: true` is implicit on `notebook:` and `bokeh_app:` commands. Translated per command type:

### `notebook:` commands

```toml
[tasks.nb]
cmd = """\
jupyter notebook report.ipynb \
--NotebookApp.default_url=/notebooks/report.ipynb \
{% if port %}--port {{ port }}{% endif %} \
{% if address %}--ip {{ address }}{% endif %} \
{% if url_prefix %}--NotebookApp.base_url={{ url_prefix }}{% endif %} \
{% if iframe_hosts %}--NotebookApp.tornado_settings={...}{% endif %} \
{% if no_browser %}--no-browser{% endif %} \
{% if use_xheaders %}--NotebookApp.trust_xheaders=True{% endif %}\
"""
args = [{ arg = "port", default = "" }, { arg = "address", default = "" }, ...]
```

### `bokeh_app:` commands

```toml
[tasks.app]
cmd = """\
bokeh serve myapp \
{% if host %}--host {{ host }}{% endif %} \
{% if port %}--port {{ port }}{% endif %} \
{% if address %}--address {{ address }}{% endif %} \
{% if url_prefix %}--prefix {{ url_prefix }}{% endif %} \
{% if not no_browser %}--show{% endif %} \
{% if use_xheaders %}--use-xheaders{% endif %}\
"""
args = [{ arg = "host", default = "" }, ...]
```

### Generic `unix:` with `supports_http_options: true`

```toml
[feature.sampleproj.tasks.dashboard]
cmd = """\
panel serve glaciers.ipynb \
{% if host %}--anaconda-project-host {{ host }}{% endif %} \
...
"""
args = [{ arg = "host", default = "" }, ...]
```

### `supports_http_options: false`

Scan the user's templated `unix:` line for HTTP Jinja vars (`{{ port }}`, `{{ host }}`, etc.) and declare pixi args only for the ones actually referenced. The cmd template is left verbatim.

### Per-tool arg counts

| shape | count | dropped |
|---|---|---|
| generic | 7 | — |
| `notebook:` | 6 | `host` (jupyter has no equivalent) |
| `bokeh_app:` | 6 | `iframe_hosts` (bokeh has no CSP) |

## `publication_info` improvements

* `info['commands'][name]['args']` — ordered list of pixi arg names. Empty for tasks without args. Lets downstream tooling drive `pixi run <task> *args` with positional values pulled from env vars or wherever.
* `info['commands'][name]['env_spec']` — resolves to an env that actually supports the task. Top-level tasks → resolved default env. Feature-scoped tasks → an env that includes the feature.
* `info['env_specs'][name]['locked']` — `True` if `pixi.lock` has an `environments[<name>]` entry. Silent `False` on any read/parse failure.
* `info['env_specs']` no longer surfaces a phantom `default` entry alongside user-declared envs.
* **Optional `project_type` argument** to force a specific manifest format. Default behavior (auto-detect, pixi-wins-when-both-present) is unchanged. Callers that need to read whichever they ask for explicitly — typically mid-conversion directories where both `pixi.toml` and `anaconda-project.yml` exist — can now pass `project_type='pixi'` or `project_type='anaconda-project'`. Asking for a format whose manifest is absent raises `FileNotFoundError`; an unrecognized value raises `ValueError`.
* **Optional `env_paths=True`** to populate `info['env_specs'][name]['path']` with the filesystem prefix for each declared environment. Off by default. For anaconda-project the path comes from each `EnvSpec.path()` — free. For pixi we shell out to `pixi info --json --manifest-path <pixi.toml>` and read `environments_info[].prefix`; failure to invoke pixi or parse its output raises `RuntimeError` since the caller explicitly opted in. When the source project is pixi *and* `env_paths=True`, the full `pixi info --json` payload is also stashed under an `_pixi` key in the result for callers that want richer pixi-specific data without making a second subprocess call.

## New `anaconda-project info` subcommand

Human-readable summary of `publication_info()` modeled on `pixi info`.

```
$ anaconda-project info --directory .
Project
------------
              Type: pixi
              Name: Attractors
       Description: A panel dashboard using datashader to interact with attractor equations.
         Directory: /Users/.../attractors

Commands
------------
           Command: dashboard  [default, http]
                  : env_spec: sampleproj
                  : cmd: panel serve attractors.ipynb {% if host %}...{% endif %} ...
                  : args: host, port, address, url_prefix, iframe_hosts, no_browser, use_xheaders

Environments
------------
       Environment: default
          Channels: https://repo.anaconda.com/pkgs/main, https://repo.anaconda.com/pkgs/r
  Dependency count: 12
      Dependencies: colorcet, datashader, fiona, geoviews, ipykernel, jupyter_bokeh, ...
            Locked: yes
```

Flags:

* `--json` — emit the `publication_info` dict as indented JSON, mirroring `pixi info --json`.
* `--env-paths` — include filesystem prefix paths for each environment. For pixi, this shells out to `pixi info --json` and (in `--json` mode) stashes the full pixi output under an `_pixi` key.
* `--project-type {pixi,anaconda-project}` — force a manifest format when both are present.

## prepare task convention (final)

Every converted project gets a `prepare` task. The body depends on context:

* **Source yml declared `downloads:`** → prepare invokes `ap_download.py` once per download.
* **No downloads** → prepare is a no-op `echo`. It still serves as the converted-project marker (downstream tooling detects converted projects by task name) and, when scoped to a non-default env's feature, as the env-selection entry point that makes `pixi run prepare` resolve to the right env.

## Cold-start perf fix

`publication_info()` was ~80x slower on `anaconda-project.yml` than on `pixi.toml` (~1150ms vs ~14ms on a small project). Profiling showed the gap was almost entirely module-import cost — paid once per process, but very visible in CLI tools and short-lived workers.

Three concurrent fixes in this PR:

1. **`current_platform()` reimplemented in pure Python** — mirrors `conda.base.context._native_subdir` (CONDA_SUBDIR override, sys.platform map, non-x86 machine table, pointer size). The previous implementation shelled out to `conda info --json` at *module-import time* (line `_default_platforms_with_current = ...`), accounting for ~1s on every `import anaconda_project`. Verified to match `conda info --json` on the dev machine; honors `CONDA_SUBDIR` so the existing parametrized non-x86 tests still pass.
2. **`default_platforms` updated** to `('linux-64', 'linux-aarch64', 'osx-arm64', 'win-64')` — the 2026 default set.
3. **Lazy imports** for heavy optional deps: `conda_pack._progress` and `tqdm` moved inside their call sites in `archiver.py`; `jinja2.Template` moved inside `_TemplateArgsTransformer.parse_and_template`. None of these are touched on the publication_info read path.

Result on a small attractors project (cold process):

| | before | after |
|---|---|---|
| anaconda-project | ~1149ms | ~180ms |
| pixi | ~14ms | ~9ms |
| anaconda-project (steady-state) | ~10ms | ~11ms |

The remaining ~140ms cold cost in dev checkouts is `_version.py` shelling out to `git describe` — versioneer writes a static `_version.py` at sdist/wheel build time, so installed distributions don't pay this cost.

## Test plan

* [x] 124 tests across `test_pixi_export.py` and `test_project_info.py` pass.
* [x] Glaciers (`panel serve` with `supports_http_options: true`): all 7 args declared; gated flags render correctly; `pixi run dashboard <host> <port> ...` works.
* [x] NYC Taxi (downloads): prepare task fetches the parquet; absent if downloads were also absent.
* [x] Multi-env with literal `default`: `default` slot commented in `[environments]`; default-feature packages folded; `env_spec` resolution names a real env.
* [x] `current_platform()` matches `conda info --json` on darwin/arm64.
* [ ] Convert a real notebook: project; verify `pixi run nb` accepts `--port` overrides correctly.
* [ ] Convert a real bokeh_app: project; verify `--show` (no-browser inverse) renders correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,11 @@ Thumbs.db
 # Project-specific
 anaconda-project-local.yml
 anaconda-project-prs/
+
+# Local AI tooling settings
+.claude/
+
+# Test artifacts produced by the local test runner
+cov.xml
+fake_project/
+fake_project.zip
--- a/anaconda_project/archiver.py
+++ b/anaconda_project/archiver.py
@@ -20,8 +20,12 @@ import tempfile
 import uuid
 import zipfile
 from io import BytesIO
-from conda_pack._progress import progressbar
-from tqdm import tqdm
+
+# `conda_pack` and `tqdm` are deferred to the call sites below
+# (`_archive_project_dir`, `_extract_files`). Importing them eagerly
+# adds tens of milliseconds to every `import anaconda_project.project`,
+# which `publication_info` and other lightweight consumers pay for no
+# reason — neither package is touched on the read path.
 
 from anaconda_project.frontend import NullFrontend, _new_error_recorder
 from anaconda_project.internal import logged_subprocess
@@ -259,6 +263,7 @@ def _leaf_infos(infos):
 
 
 def _write_tar(archive_root_name, infos, filename, compression, packed_envs, frontend):
+    from conda_pack._progress import progressbar
     if compression is None:
         compression = ""
     else:
@@ -294,6 +299,7 @@ def _write_tar(archive_root_name, infos, filename, compression, packed_envs, fro
 
 
 def _write_zip(archive_root_name, infos, filename, packed_envs, frontend):
+    from conda_pack._progress import progressbar
     with zipfile.ZipFile(filename, 'w') as zf:
         for info in _leaf_infos(infos):
             arcname = os.path.join(archive_root_name, info.relative_path)
@@ -446,6 +452,7 @@ def _extractall_chmod(zf, destination):
 
 
 def _extract_files_zip(zip_path, src_and_dest, frontend):
+    from tqdm import tqdm
     # the zipfile API has no way to extract to a filename of
     # our choice, so we have to unpack to a temporary location,
     # then copy those files over.
@@ -472,6 +479,7 @@ def _extract_files_zip(zip_path, src_and_dest, frontend):
 
 
 def _extract_files_tar(tar_path, src_and_dest, frontend):
+    from tqdm import tqdm
     with tarfile.open(tar_path, mode='r') as tf:
         if isinstance(frontend.underlying, NullFrontend):
             src_and_dest = tqdm(src_and_dest, desc='Extract ')
--- a/anaconda_project/internal/cli/info.py
+++ b/anaconda_project/internal/cli/info.py
@@ -1,0 +1,144 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2026, Anaconda, Inc. All rights reserved.
+#
+# Licensed under the terms of the BSD 3-Clause License.
+# The full license is in the file LICENSE.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+"""`anaconda-project info` — human-readable summary of publication_info."""
+from __future__ import absolute_import, print_function
+
+import json
+import sys
+
+from anaconda_project.project_info import (
+    PROJECT_TYPE_PIXI,
+    PROJECT_TYPE_KEY,
+    publication_info,
+)
+
+
+def _format_section(title):
+    return '{}\n{}'.format(title, '-' * 12)
+
+
+def _format_label_value(label, value, width=18):
+    return '{:>{w}}: {}'.format(label, value, w=width)
+
+
+def _format_label_continuation(value, width=18):
+    return '{:>{w}}: {}'.format('', value, w=width)
+
+
+def _format_list(label, items, width=18):
+    """Format a label with a list of values, one per line, joined by ', '
+    on overflow. Mirrors `pixi info`'s layout for `Dependencies:`."""
+    if not items:
+        return _format_label_value(label, '', width=width)
+    return _format_label_value(label, ', '.join(items), width=width)
+
+
+def _print_text(info, project_dir):
+    """Render info as text, modeled on `pixi info`."""
+    project_type = info.get(PROJECT_TYPE_KEY, '<unknown>')
+    print(_format_section('Project'))
+    print(_format_label_value('Type', project_type))
+    print(_format_label_value('Name', info.get('name', '')))
+    if info.get('description'):
+        print(_format_label_value('Description', info['description']))
+    print(_format_label_value('Directory', project_dir))
+    print()
+
+    commands = info.get('commands') or {}
+    if commands:
+        print(_format_section('Commands'))
+        for name, cmd in commands.items():
+            tags = []
+            if cmd.get('default'):
+                tags.append('default')
+            if cmd.get('notebook'):
+                tags.append('notebook')
+            elif cmd.get('bokeh_app'):
+                tags.append('bokeh_app')
+            if cmd.get('supports_http_options'):
+                tags.append('http')
+            header = name
+            if tags:
+                header = '{}  [{}]'.format(name, ', '.join(tags))
+            print(_format_label_value('Command', header))
+            if cmd.get('env_spec'):
+                print(_format_label_continuation('env_spec: {}'.format(cmd['env_spec'])))
+            run_line = cmd.get('unix') or cmd.get('windows') or cmd.get('notebook') or cmd.get('bokeh_app') or ''
+            if run_line:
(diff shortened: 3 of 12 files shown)


A-Aron-T:
Tested the full integration path end-to-end:

1. **Export** — `supports_http_options` commands produce pixi tasks with `{% if var %}...{% endif %}` gates and `args = [{ arg = "host", default = "" }, ...]` declarations. Verified with Panel (`panel serve`), generic flask-style, `bokeh_app:`, and `notebook:` commands.

2. **pixi task runner** — Confirmed pixi 0.67.2 natively handles the `{% if %}` and `{{ var }}` template syntax in task `cmd` strings. Empty string args → falsy → flag omitted. Non-empty → truthy → flag emitted. Boolean-style flags (no value, just presence) work correctly.

3. **All three command shapes verified:**
   - Generic (`supports_http_options: true`): 7 args, `--anaconda-project-*` flags
   - `bokeh_app:`: 6 args (no `iframe_hosts`), bokeh-native flags (`--host`, `--port`, `--address`, `--prefix`). Inverted `{% if not no_browser %}--show{% endif %}` gate confirmed — empty = show browser, truthy = headless.
   - `notebook:`: 6 args (no `host`), Jupyter-native flags (`--ip`, `--NotebookApp.*`)

4. **Platform integration (`start_http_server.py`)** — Our launcher calls `pixi run --locked <command> <positional args...>` where positions match the declared `args` order. Tested that `publication_info()` returns args in declaration order.

   ⚠️ **Found a downstream integration bug in our `start_http_server.py` (line 128)**: the lookup does `standard_args.get(f'--anaconda-project-{arg}')` but pixi arg names use underscores (`url_prefix`, `iframe_hosts`, `no_browser`, `use_xheaders`) while `standard_args` keys use hyphens (`--anaconda-project-url-prefix`). Result: 4 of 7 args silently resolve to empty string. Fix is `arg.replace('_', '-')` in the lookup. We'll fix this on our side before these PRs merge — not a blocker for this PR.

5. **`anaconda-project info` subcommand** — Clean structured output in both human and `--json` mode. Correctly detects project type, tags commands with `[default, http]` / `[notebook, http]` labels. `--env-paths` resolves prefix locations via `pixi info --json`. All 10 info tests pass.

6. **`publication_info` improvements** — Confirmed `args` list, `env_spec` resolution, `project_type` forcing, `locked` detection, and `env_paths` all work. Cold-start at ~120ms (vs ~1149ms before the perf fix). 75 project_info tests pass.

**CI note:** The `test_main_help` test in `test_main.py` needs `'info'` added to the `all_subcommands` tuple (line 20) and a corresponding help text line. This is what's causing the CI failure in the rollup (#434) — trivial fix.

One callout: the `{% if %}` / `{{ }}` syntax in pixi tasks isn't prominently documented in pixi's own docs. It works today (deno_task_shell supports it), but might be worth a brief mention in the export docs so future maintainers know it's intentional, not accidental.

Looks good — this is the largest PR in the set and it holds up well.

mcg1969:
Thanks for the thorough end-to-end test — appreciate the cross-confirmation that pixi 0.67.2 actually parses the templated `cmd` strings as we expected.

Two items addressed:

* **CI fix** — pushed [646a2ed](https://github.com/anaconda/anaconda-project/commit/646a2ed) on this branch: added `'info'` to `all_subcommands` and the corresponding wrapped help line in `test_main.py`. That unblocks `test_main_help` / `test_main_help_via_entry_point` on the rollup. Verified locally — 7 passed.
* **Doc note on the `{% if %}` / `{{ var }}` syntax** — landed on `feature/docs-update` ([b0c2764](https://github.com/anaconda/anaconda-project/commit/b0c2764)) since this PR's branch doesn't carry the pixi-support page. Quick correction worth noting: pixi renders task `cmd` through MiniJinja *before* handing the rendered string to `deno_task_shell` for execution — the templating and the shell are two separate stages. The doc note reflects that.

On your `start_http_server.py` finding (the `arg.replace('_', '-')` bug at line 128) — good catch, no action needed on our side. Let us know if you'd like the `args` keys flipped to hyphens in the converter instead; happy to discuss, but underscores match pixi's own convention for arg names.

PR #431, opened 2026-05-15
docs: project status, pixi support page, and modern feedback channels

### LINKED ISSUE #432: Drop pkg_resources fallback in internal.plugins; use importlib.metadata
## Symptom

`anaconda_project.internal.plugins` falls back to `pkg_resources.iter_entry_points` when the third-party `entrypoints` package is unavailable. On `setuptools >=81`, `pkg_resources` is no longer installed by default, so the fallback path raises `ModuleNotFoundError: No module named 'pkg_resources'` at import time.

```python
try:
    from entrypoints import get_group_named
except ImportError:  # py 2.7
    from pkg_resources import iter_entry_points
    ...
```

## Why this matters downstream

`anaconda-project-launcher` (and likely other consumers) hits this at server-extension load time. The launcher's serverextension imports `anaconda_project.project`, which transitively imports `anaconda_project.internal.plugins`. On a fresh setuptools-81+ environment without `entrypoints`, the launcher's extension fails to load with no obvious connection to setuptools.

We worked around it in `anaconda-project-launcher` by adding `entrypoints` to our run requirements. That works but is a workaround.

## Suggested fix

Replace the `pkg_resources` fallback with `importlib.metadata.entry_points`, which has been in the stdlib since Python 3.8:

```python
try:
    from entrypoints import get_group_named
except ImportError:
    from importlib.metadata import entry_points
    def get_group_named(group_name):
        return {ep.name: ep for ep in entry_points(group=group_name)}
```

`importlib.metadata.entry_points()` returns a `SelectableGroups` in 3.10+ that supports the `group=` keyword. For 3.8/3.9 compatibility, the `entry_points()` call returns a dict-of-lists keyed by group name; a small adapter handles both.

The third-party `entrypoints` fast-path can stay or be removed — it's no longer materially faster than the stdlib equivalent on modern Pythons.

## Reference

This was discussed during anaconda-project-launcher PR work; the workaround landed in https://github.com/anaconda/anaconda-project-launcher (added `entrypoints` to run deps).

mcg1969:
## Summary

Stacked on #435 (which is itself stacked on #430). Documentation refresh and infrastructure work to get the docs building cleanly on Read the Docs again, plus a new pixi support page covering the conversion features in #428–#430 and the export flags / public API in #435.

## Read the Docs revival

Three CI/config commits to get RTD building again:

* **`.readthedocs.yaml`** — RTD now requires a config file in the repo root or it fails to detect the project. Use the v2 schema with `miniforge3-latest`, our existing `docs/environment.yml`, and `docs/source/conf.py`.
* **`docs/environment.yml`** — drop `defaults` from the channel list (RTD can't accept the Anaconda ToS gate, and the solver was silently dropping `pydata-sphinx-theme`); unpin the theme so we get whatever conda-forge has shipped (currently 0.17.1).
* **`conf.py` html_theme guard** — fix the inverted `not on_rtd` check that left the theme at sphinx's default (alabaster) on RTD, then crashed because the pydata-only theme options below assumed pydata was loaded.

## Documentation refresh

* **New "Project status" section** on the front page acknowledging that anaconda-project has not been actively developed in some time and is largely superseded. Inline links to the four alternatives:
  * [conda-project](https://conda-incubator.github.io/conda-project/)
  * [pixi](https://pixi.sh/)
  * [uv](https://docs.astral.sh/uv/)
  * [conda-workspaces](https://github.com/conda-incubator/conda-workspaces)

  Notes that recent activity is centered on conversion to pixi / conda-workspaces.
* **New `pixi-support.rst` page** under its own caption in the docs nav, alongside Get started, User guide, and Reference. Covers:
  * Conversion conventions for env layout, channel handling, and command translation (the work in #428–#430).
  * The `prepare` task contract — when it's emitted and what it does — including the new behavior under `--use-default` (no-op prepare is dropped; only emitted when there are real downloads to fetch).
  * The HTTP-options dispatch table per command type (`unix:`, `notebook:`, `bokeh_app:`).
  * The `--use-default` and `--add-current-platform` flags from #435: what they do, the selection rules, when they're no-ops, and the recommendation/confirmation messages the CLI prints around them.
  * A "Programmatic export API" section documenting the hooks downstream consumers depend on: `project_ops.export_pixi` (returns `PixiExportStatus`), `project_ops.preview_pixi_export` (returns the four-key preview dict), and the public pickers `default_rename_target` / `current_platform_addition_target`.
  * The `info` command and the `publication_info` schema added in #429–#430.
  * The limitations that can't be translated faithfully (services, prompted variables).
* **Help and support page rewrite** — adds a "Project scope" callout pointing readers to pixi / conda-workspaces / uv for new project work, and replaces the stale `documentation@continuum.io` email with a redirect to the GitHub Issue Tracker. Anaconda, Inc. (formerly Continuum Analytics) hasn't monitored that address in years.
* **Disambiguate "Project"** — the bare word was ambiguous between the generic noun ("a project directory") and the tool itself. Walk the docs and replace bare `Project` with `anaconda-project` whenever the tool was meant. Brand-form `Anaconda Project` references and genuinely generic uses left alone. Drop the orphan "Stability" section and refresh the copyright header. Affects `index.rst`, `getting-started.rst`, `user-guide/concepts.rst`, and three task pages.
* **Install page rewrite** — anaconda-project hasn't been bundled with the Anaconda installer since 2022.05; the page still claimed it was and walked users through `conda list` to look for it. Replace with a short page covering install from defaults and conda-forge, the verify command, and a Miniconda pointer for users who don't yet have conda. (-73 / +11 lines.)

Fixes #432

## Test plan

* [x] No tests changed — existing tests unaffected.
* [x] Local sphinx build with `docs/environment.yml` passes under `-W` (warnings-as-errors). All four inline links resolve in the rendered HTML; the new section appears in the page TOC between the intro and Quick Start.
* [x] Verified RTD config locally with the same env file using a clean conda env.
* [ ] CI sphinx build passes.
* [ ] Read the Docs build passes after merge.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,20 @@
+# Read the Docs configuration. See:
+# https://docs.readthedocs.com/platform/stable/config-file/v2.html
+#
+# Read the Docs requires this file (in the repository root) to build
+# the docs. Sphinx config and dependencies live under docs/.
+
+version: 2
+
+build:
+  os: ubuntu-24.04
+  tools:
+    # miniforge3 supplies conda + the conda-forge channel by default,
+    # which matches the resolution we get locally with the env file.
+    python: "miniforge3-latest"
+
+conda:
+  environment: docs/environment.yml
+
+sphinx:
+  configuration: docs/source/conf.py
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -1,9 +1,11 @@
 name: anaconda-project-docs
 dependencies:
 - sphinx
-- pydata-sphinx-theme=0.14.4
+- pydata-sphinx-theme
 - sphinx-copybutton
 - sphinx-notfound-page
 channels:
-- defaults
+# `defaults` is intentionally omitted: Read the Docs build images
+# can't accept the Anaconda ToS gate that channel now requires, and
+# every package listed above is available on conda-forge.
 - conda-forge
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -8,4 +8,10 @@ channels:
 # `defaults` is intentionally omitted: Read the Docs build images
 # can't accept the Anaconda ToS gate that channel now requires, and
 # every package listed above is available on conda-forge.
+#
+# This applies only to the docs build environment. Projects exported
+# via `anaconda-project export-pixi` still resolve against the user's
+# configured `default_channels` (typically `pkgs/main` + `pkgs/r`);
+# the conda-forge-only choice here is a build-tooling decision, not
+# a project-level default.
 - conda-forge
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,8 +13,6 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import os
-
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
@@ -126,10 +124,12 @@ todo_include_todos = False
 # a list of builtin themes.
 #
 
-on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
-
-if not on_rtd:  # only import and set the theme if we're building docs locally
-    html_theme = 'pydata_sphinx_theme'
+# Always set the theme. The previous `if not on_rtd:` guard skipped this
+# when building on Read the Docs (where READTHEDOCS=True), back when RTD
+# injected its own theme. RTD no longer does that — without the explicit
+# assignment sphinx falls back to alabaster, which doesn't understand the
+# pydata-specific options below and fails to render.
+html_theme = 'pydata_sphinx_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further. The documentation for the current theme is found at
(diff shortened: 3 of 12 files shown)


A-Aron-T:
Docs look good. The RTD revival (`.readthedocs.yaml` + conda-forge-only docs env) is the right approach — avoids the Anaconda ToS gate on RTD builders.

The pixi support page covering #428–#430 and #435 features will be useful for users discovering the export path. Nice that it documents the `info` subcommand and `publication_info` options as well.

One note: the docs env uses conda-forge only (which is fine for build tooling — sphinx, theme, etc.), but worth a comment somewhere that the _exported projects_ still resolve against `pkgs/main` per #429's channel defaults. Just to avoid confusion if someone reads the docs env.yml and thinks conda-forge is the project-level default.

mcg1969:
Thanks. Both items addressed in [b0c2764](https://github.com/anaconda/anaconda-project/commit/b0c2764) on this branch:

* **`docs/environment.yml`** — added a comment explaining the conda-forge-only choice is for the docs build environment specifically, and that exported projects still resolve against the user's configured `default_channels` (typically `pkgs/main` + `pkgs/r` per #429).
* **Pixi template-syntax note** — added to the HTTP options section of `pixi-support.rst`, explaining that the `{% if %}` / `{{ var }}` syntax is intentional. Pixi renders task `cmd` through MiniJinja before deno_task_shell ever sees it; the feature is supported but not prominently documented upstream.

PR #435, opened 2026-05-17
feat(export-pixi): --use-default and --add-current-platform flags

mcg1969:
## Summary

Two related export-pixi flags for cleaner pixi.toml output, plus the public API hooks downstream consumers asked for.

### `--use-default`

* Renames a promoted env_spec to `default` on export, collapsing `[feature.{name}.*]` blocks into pixi's top-level `[dependencies]` / `[tasks.X]` / `[tasks.prepare]` sections.
* Selection rule: the env_spec attached to the project's default command (what a bare `anaconda-project run` would invoke); if there are no commands, the first env_spec declared. No-op when an env_spec literally named `default` already exists.
* Drops two low-signal `# converted from notebook/bokeh_app command` comments from generated tasks.
* Drops the no-op `prepare` task. The unflagged export emits an `echo` placeholder under `[feature.{name}.tasks.prepare]` so `pixi run prepare` resolves to the project's intended env even when its name isn't `default`. Under `--use-default` the promoted env *is* the implicit default, so the placeholder loses its purpose; `prepare` is only emitted when there are real `downloads:` to fetch.

### `--add-current-platform`

* Widens the converted manifest's `platforms` list to include the host's conda subdir (e.g. `osx-arm64`) when it isn't already declared. Pixi rejects envs that don't list the host platform; anaconda-project is more forgiving, so this prevents a pixi install error after conversion.
* Off by default; the exporter does not silently mutate the user's platforms list.

### CLI behavior (mirrored across both flags)

* When the user omits the flag and the project would benefit, the CLI prints a recommendation naming what would change.
* When the user passes the flag and something actually changed, the CLI prints a confirmation. No-op (e.g. flag passed but the project already has `default` / already lists the host platform) is silent.

### Public API for downstream consumers

A separate commit responds to upstream-consumer feedback by exposing stable hooks instead of forcing them through underscored internals:

* `default_rename_target(project)` and `current_platform_addition_target(project)` — public pickers; return what would change, or `None` for no-op.
* `PixiExportStatus` (returned by `project_ops.export_pixi`) carries `default_rename_from` and `current_platform_added`. Lets a launcher show "Renamed env X → default" / "Added platform Y" without re-querying.
* `project_ops.preview_pixi_export(project, use_default=False, add_current_platform=False)` returns `{pixi_toml, default_rename_from, current_platform_addition_target, warnings}` — a single call so multiple frontends (CLI, launcher, IDE plugin) can render the same dialog without divergence.

`extract_warnings(content)` is factored out so both `export_pixi` and `preview_pixi_export` parse the leading `# WARNING:` block the same way.

## Commits

1. `feat(export-pixi): --use-default flag to rename promoted env to "default"`
2. `feat(export-pixi): public API for downstream pixi-conversion consumers`
3. `feat(export-pixi): --add-current-platform flag`
4. `refactor(export-pixi): omit no-op prepare task under --use-default`

## Test plan

- [x] `pytest anaconda_project/internal/test/test_pixi_export.py` — 96 passing (23 new tests across `TestUseDefault`, `TestExtractWarnings`, `TestPublicAPI`, and `TestAddCurrentPlatform` covering pickers, single-env collapse, multi-env promotion, no-op semantics, status fields, preview shape, current-platform addition with monkeypatched `current_platform()`, and prepare-task omission under `--use-default`).
- [x] End-to-end CLI smoke tests: single-env, multi-env with default command, project with `default` already present, project missing host platform, both flags together, project with downloads under `--use-default`.

User-facing documentation lives in #431 (the rebased docs branch on top of this one).

### DIFF
--- a/anaconda_project/internal/cli/main.py
+++ b/anaconda_project/internal/cli/main.py
@@ -414,6 +414,14 @@ def _parse_args_and_run_subcommand(argv):
                               "'default' on export. Collapses [feature.{name}.*] blocks "
                               "into top-level [dependencies] / [tasks.X] for a cleaner "
                               "pixi.toml."))
+    preset.add_argument('--add-current-platform',
+                        action='store_true',
+                        default=False,
+                        help=("If the host's conda subdir (e.g. osx-arm64) isn't already in "
+                              "the project's platforms list, add it on export. Pixi rejects "
+                              "envs that don't list the host platform; anaconda-project is "
+                              "more forgiving, so this prevents a pixi install error after "
+                              "conversion."))
     preset.set_defaults(main=pixi_commands.main_export_pixi)
 
     preset = subparsers.add_parser(
--- a/anaconda_project/internal/cli/main.py
+++ b/anaconda_project/internal/cli/main.py
@@ -406,6 +406,14 @@ def _parse_args_and_run_subcommand(argv):
     preset = subparsers.add_parser('export-pixi', help="Export the project as a pixi.toml file")
     add_directory_arg(preset)
     preset.add_argument('filename', metavar='PIXI_TOML_FILE', nargs='?', default='pixi.toml')
+    preset.add_argument('--use-default',
+                        action='store_true',
+                        default=False,
+                        help=("If the project has no env_spec named 'default', rename the "
+                              "default-command's env_spec (or the first one declared) to "
+                              "'default' on export. Collapses [feature.{name}.*] blocks "
+                              "into top-level [dependencies] / [tasks.X] for a cleaner "
+                              "pixi.toml."))
     preset.set_defaults(main=pixi_commands.main_export_pixi)
 
     preset = subparsers.add_parser(
--- a/anaconda_project/internal/cli/pixi_commands.py
+++ b/anaconda_project/internal/cli/pixi_commands.py
@@ -13,7 +13,7 @@ from anaconda_project.internal.cli import console_utils
 from anaconda_project import project_ops
 
 
-def export_pixi(project_dir, filename):
+def export_pixi(project_dir, filename, use_default=False):
     """Export the project as a pixi.toml file."""
     import os
     project = load_project(project_dir)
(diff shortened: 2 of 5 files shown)


A-Aron-T:
Tested locally. All three features verified:

**`--use-default`** — Confirmed it renames the promoted env_spec to pixi's implicit "default", collapsing `[feature.{name}.*]` blocks into top-level `[dependencies]` / `[tasks]`. Selection rule (default command's env_spec) is sensible. The `prepare` task is correctly suppressed (no longer needed as an env-selection entry point when the env _is_ the default — only emitted when there are real `downloads:` to fetch).

**`--add-current-platform`** — Adds the host platform to the exported `platforms` list. Verified it adds `osx-arm64` to a `linux-64`-only project and prints the confirmation message. Good that it's off-by-default and prints a recommendation when omitted.

**Public API** — `preview_pixi_export`, `default_rename_target`, and `current_platform_addition_target` all importable and functional:
- `preview_pixi_export(project, use_default=True)` returns `{'pixi_toml': ..., 'default_rename_from': ..., 'current_platform_addition_target': ..., 'warnings': ...}` — exactly the shape needed for preview dialogs.
- `PixiExportStatus` (returned by `project_ops.export_pixi`) carries `default_rename_from` and `current_platform_added` fields so callers can report what changed without re-querying.
- Both flags together work correctly in a single call.

One question: `preview_pixi_export` requires a loaded `Project` object. For downstream CLI consumers that only have a directory path, the `internal.cli.pixi_commands.export_pixi(project_dir, ...)` wrapper is the simpler entry point — is that intended to be stable, or should we stick to loading `Project` ourselves?

Looks good.

mcg1969:
Thanks for testing the public API surface so thoroughly.

On your question about `internal.cli.pixi_commands.export_pixi(project_dir, ...)`: short answer — that one is not currently a stable public entry point. The `internal/` namespace is treated as private per `anaconda_project/api.py`'s docstring ("anything with an underscore prefix or inside `anaconda_project.internal` is considered private"), so I'd recommend continuing to load `Project` yourself for now.

That said, your use case is reasonable — a downstream caller that has a directory and wants the same "load → check problems → export → print recommendations" sequence the CLI runs. If that's a recurring pattern for the launcher, the right move is to promote a `project_dir`-taking helper into `anaconda_project.api` (probably as `AnacondaProject.export_pixi(directory_path, filename, ...)` returning the same `PixiExportStatus`). Happy to do that as a follow-up — just say the word and I'll open a separate PR. Doesn't need to block this one.

A-Aron-T:
Thanks for clarifying. We're good without the `project_dir`-taking wrapper — our main downstream path is `publication_info(project_dir)` which already has the simple interface we need. The `preview_pixi_export(project)` signature works for our promote dialog since we have a loaded Project in that context.

No need for a follow-up PR on this. Appreciate the offer though.

PR #436, opened 2026-05-29
AENT-8833 / AENT-8838: export-pixi shell-quote fix + default_channels= seam

A-Aron-T:
## Summary

`export-pixi` could produce a `pixi.toml` task that runs attacker-controlled shell commands. For `notebook:` and `bokeh_app:` commands, `_command_to_task` synthesized the deno_task_shell command string by interpolating the path directly:

```python
raw_cmd = 'jupyter notebook {}'.format(command.notebook)
raw_cmd = 'bokeh serve {}'.format(command.bokeh_app)
```

The result was escaped only by `_toml_string`, which handles TOML syntax but **not** shell metacharacters (`;`, `|`, `$(...)`, backtick). So a project whose `notebook:`/`bokeh_app:` field contains those would generate a pixi task that executes the injected command on the next `pixi run`.

anaconda-project's own runner execs these as an **argv list** (`['jupyter-notebook', path]`, no shell — `project_commands.py`), where the path is inert. The converter was *introducing* a shell-injection that does not exist in the native run path.

### Why it matters

The project author is usually the person who runs the project, so for self-authored projects this is low impact. But `export-pixi` is explicitly used to convert **shared / sample / gallery** projects (e.g. the sample-projects → pixi conversion), where author ≠ runner — a supply-chain-flavored RCE: convert someone else's project, then `pixi run` executes their payload.

## Fix

Route both paths through the existing `_shell_quote` helper (escapes `" \ $` and backtick). Clean filenames are unaffected — they just gain surrounding quotes, which deno_task_shell treats identically.

## Tests

- New: notebook/bokeh metacharacter-neutralization cases — **mutation-verified** (both fail if the quoting is removed).
- New: direct `_shell_quote` escaping coverage (the helper was previously untested).
- Updated: the two happy-path notebook/bokeh assertions for the now-quoted form.
- `106 passed` locally (was 96 + happy-path edits + 6 new).

## Notes

- Found via an adversarial correctness audit of the merged pixi pipeline; verified against the actual code + the native runner's argv behavior.
- Touches `pixi_export.py`, which has had recent activity — opening as **draft** for now; not requesting review yet.

---

## Added: AENT-8838 — `default_channels=` seam (commit 702add8)

`export_pixi_toml` (and the `export_pixi` / `preview_pixi_export` wrappers) gain an optional `default_channels=` parameter. When supplied, it expands the `defaults` meta-channel and serves as the empty-channels fallback, and the `conda config --show default_channels` subprocess is **skipped entirely**.

This is the seam the editor pixi-launcher needs: it can resolve channels itself from the session `.condarc` and call this canonical converter with **no conda subprocess**, so the editor's standalone `standalone_export.py` fork (which exists only to dodge the conda-info hang under Rosetta) can later be deleted.

- `default_channels=None` (default) → unchanged behavior (local conda lookup).
- New `TestDefaultChannelsParam` (6 tests); the "conda must not be called" assertions are **mutation-verified** (fail if the param is ignored or the wrappers stop threading it).

Bundling with the injection fix because both touch `pixi_export.py` and ship in one ae-dev release.

### DIFF
--- a/anaconda_project/internal/pixi_export.py
+++ b/anaconda_project/internal/pixi_export.py
@@ -647,9 +647,15 @@ def _command_to_task(command, declared_vars):
         raw_cmd = command.unix_shell_commandline
         raw_forms.append(raw_cmd)
     elif command.notebook is not None:
-        raw_cmd = 'jupyter notebook {}'.format(command.notebook)
+        # Quote the path: anaconda-project runs notebooks as an argv list
+        # (['jupyter-notebook', path], no shell), so a notebook field
+        # containing shell metacharacters is inert there. Here we synthesize a
+        # deno_task_shell command string, so an unquoted path would let
+        # metacharacters (;, |, $(), backtick) break out and execute when a
+        # user later runs `pixi run`. _shell_quote neutralizes them.
+        raw_cmd = 'jupyter notebook {}'.format(_shell_quote(command.notebook))
     elif command.bokeh_app is not None:
-        raw_cmd = 'bokeh serve {}'.format(command.bokeh_app)
+        raw_cmd = 'bokeh serve {}'.format(_shell_quote(command.bokeh_app))
     elif command.windows_cmd_commandline:
         # No unix variant — normalize the windows form so it runs under
         # deno_task_shell on every platform pixi targets.
--- a/anaconda_project/internal/pixi_export.py
+++ b/anaconda_project/internal/pixi_export.py
@@ -779,7 +779,7 @@ def current_platform_addition_target(project):
     return None if here in declared else here
 
 
-def export_pixi_toml(project, use_default=False, add_current_platform=False):
+def export_pixi_toml(project, use_default=False, add_current_platform=False, default_channels=None):
     """Convert an anaconda-project Project to pixi.toml content.
 
     Args:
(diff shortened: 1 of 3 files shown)


mcg1969 (approved):
## Review: LGTM — approving

Reviewed the full diff, the surrounding code, and ran the suite locally (98 passed on the branch base; CI shows 106 green across the matrix). Both changes are correct, the security framing is accurate rather than inflated, and the test claims hold up.

### Injection fix (AENT-8833)

The vulnerability is real and the analysis is precise:

* `notebook:`/`bokeh_app:` fields were interpolated directly into a `deno_task_shell` command string, then escaped only by `_toml_string`, which handles TOML syntax — not shell metacharacters.
* The claim that anaconda-project's *native* runner is safe (argv list, no shell — `project_commands.py:486-494`) while the converter *introduces* the injection is correct. This is the right thing to emphasize: it matters specifically for the shared/gallery conversion path where author ≠ runner.
* `_shell_quote` escapes exactly the four characters special inside a `deno_task_shell` double-quoted token (`" \ $` backtick) and leaves `; | &` as inert data inside the quotes — the correct set. The double-layer escaping (shell_quote → toml_string → TOML parse → deno parse) is sound; the bokeh test pinning the doubled `app\\$(...)` form confirms it survives both layers.

The mutation-verification on the two neutralization tests is good practice — they genuinely fail if the quoting is removed.

On the ordering change (`_shell_quote` now runs *before* `_translate_command_env_vars`): I traced this through the native runner and it is a non-issue — in fact the new behavior is *more* correct. The native runner treats the notebook/bokeh field as a literal relative path via `os.path.join(environ['PROJECT_DIR'], self.notebook)` with no `os.path.expandvars`, so env-var expansion in those fields is not a supported feature. The old converter's env-var translation over those paths was inventing a feature that doesn't exist in the run path; quoting the `$` to a literal keeps the path inert, matching native semantics. The `unix_shell_commandline` path — which legitimately runs under a shell and does need env-var translation — is untouched, which is correct.

### `default_channels=` seam (AENT-8838)

Clean and correctly threaded through both `project_ops` wrappers. The refactor is behavior-preserving for the `default_channels=None` case (every existing caller): the new guarded assignment is identical to the old conditional. The "conda must not be called" assertions via the exploding `boom()` fixture are the right way to prove the subprocess is actually skipped.

### Approving

No blockers. 🤖 reviewed with [Claude Code](https://claude.com/claude-code)

PR #439, opened 2026-06-25
AENT-8881: no-solver anaconda-project-lock → pixi.lock translator (Items 1/2/4/5, live-verified)

A-Aron-T:
## Summary

Adds a no-solver `anaconda-project-lock.yml` → `pixi.lock` translator so a converted pixi project installs the *exact* packages it was locked with via `pixi install --locked`, instead of `pixi install` re-solving (which provably does NOT reproduce complex environments). Live-verified on ROSA (see "Live verification" below). Ready for review.

**Scope correction:** earlier revisions of this PR description said "Item 1 only." That's no longer accurate — this branch has accumulated Items 1, 2, 4, and 5 (Item 3, foreign-subdir handling, was folded into Item 2's enrichment). Only the **consumer wiring** (calling this from the launcher/CLI convert path) and the live census E2E are still pending, tracked as follow-ups on AENT-8881. See "What's in this PR" below for the real scope.

## Why (proven, not theoretical)
Spike on the holoviz `census` project (251 pkgs): **both** loose re-solve AND fully-pinned re-solve FAIL — any re-solve re-evaluates transitive constraints against *today's* repodata under strict channel priority, and metadata/channel drift after lock-time makes the solver reject a combination that worked. Only direct lock translation + `--locked` (never invoking the solver) reproduces it. Customer model-dev environments (bank/high-profile) are in scope, so this is a real reproducibility risk.

## What's in this PR
`anaconda_project/internal/pixi_lock.py`:
- `parse_locked_specs(lock_set, platform)` → exact `(name, version, build)` closure, reusing `CondaLockSet.package_specs_for_platform` (merges all/unix/<platform> buckets → cross-platform **noarch** included). Fails loud if any spec isn't pinned to exact `version=build`.
- `assert_no_pip_packages(lock_set)` → fails fast on a pip bucket (v1 is conda-only).
- `enrich_locked_packages(locked, channels, subdir, query=None)` → exact-match repodata enrichment (url/sha256/md5/depends) via conda's `SubdirData.query_all`, no solver, no substitution. Distinguishes a transport failure (`MirrorUnavailableError`) from an authoritatively-absent build (`BuildNotFoundError`) so a flaky mirror is never mistaken for "build is gone."
- `build_pixi_lock(...)` / `write_pixi_lock(...)` → assembles a pixi.lock v6 document and writes it atomically (temp file + `os.replace`).
- `translate_lock_set(lock_set, channels, path, ...)` → end-to-end entry point tying it all together.
- `_resolve_subdir_data()` → guards the conda private-API seam (`SubdirData.query_all`) with a typed `CondaIndexAPIError` + a CI canary, since this is the one place anaconda-project imports conda as a library rather than shelling out to it.

## Design decisions (settled via adversarial review — see AENT-8881)
- **strict-only v1.** A `best-effort` nearest-build mode was deliberately deferred: doing it correctly needs cross-package constraint re-derivation = a solver by another name.
- **Trust-on-first-use provenance.** `conda_api.ParsedSpec` carries **no per-package hash** — there is no lock-time hash to validate the enriched `sha256` against; integrity rests on trusted/pinned channels (the airgap mirror in AE5).
- **No re-fork** — reuses `CondaLockSet` + `conda_api.parse_spec` (the standalone_export / AENT-8833 lesson).

## Follow-up fixes (this push)
A fresh adversarial re-review after the branch had accumulated Items 1/2/4/5 found 5 spec shapes that slipped past the strict fail-loud gate the whole feature depends on:
1. **Wildcard/partial build strings accepted as exact pins.** `conda_api.parse_spec`'s wildcard filter (`|`, `*`, `,`) covers `exact_version` but not `exact_build_string`, so `numpy=1.0=*` or `numpy=1.0=py37*` were treated as fully resolved. Now explicitly rejected in `parse_locked_specs`.
2. **Nested pip packages escaped `assert_no_pip_packages`.** It only checked the top-level `"pip"` platform key, missing a `{"pip": [...]}` dict nested inside another platform's spec list (the environment.yml-style shape). Both shapes are now checked, and the nested case in `parse_locked_specs` raises a typed `LockTranslationError` instead of leaking `conda_api.parse_spec`'s raw `TypeError`.
3. **Disabled lock set surfaced an opaque `AssertionError`.** `CondaLockSet.package_specs_for_platform` has a bare `assert self.enabled`; `parse_locked_specs` now checks `lock_set.disabled` up front.
4. **Duplicate/conflicting pins within one platform bucket passed through unresolved**, deferring the failure to a confusing pixi-side rejection instead of a clear translate-time error.
5. **Zero-platform / zero-package lock sets silently wrote a valid-but-empty pixi.lock** (install nothing). `translate_lock_set` now refuses.

## Tests / verification
**30 offline unit tests** (no network, no solver, no conda index) — up from the original 6, covering all of the above plus the pre-existing enrichment/build/write/end-to-end coverage. All pass; flake8 clean.

## Live verification (ROSA)

Built this branch's `anaconda-project` conda package, overlaid it into a real editor image, pushed to a real ROSA cluster, and ran `translate_lock_set` from inside a real JupyterLab session against the **real** `repo.anaconda.com/pkgs/main` mirror — not an offline fixture:

- Hand-built a `CondaLockSet` with 3 real, exactly-pinned specs (`python=3.12.13=h4d16e0c_1`, `readline=8.3=hc2a1206_0`, `sqlite=3.53.2=h795bf6d_0`) and called `translate_lock_set` against the live channel.
- Result: a correct pixi.lock v6 document — real download URLs, real `sha256`/`md5` from the channel's actual current repodata, real `depends` edges, deduped top-level `packages` list. Exactly the shape the design promises.
- This also incidentally exercised the `CondaIndexAPIError`/`MirrorUnavailableError` guard for real: run first from the editor's `lab_launch` env (no `conda` library installed there, only the `pixi` CLI) — it correctly raised the typed error instead of an opaque `ModuleNotFoundError`, confirming the private-API guard fires correctly on a genuinely conda-less environment. Re-ran from an env with `conda` importable to get the full enrichment path.
- This is the same overlay/session used to verify anaconda-project-launcher#343 (AENT-8884), so the fix is confirmed present in the actual editor image build path (`editor/lab_launch/pixi.toml`), not just in this repo's own test suite.

### DIFF
--- a/anaconda_project/internal/pixi_lock.py
+++ b/anaconda_project/internal/pixi_lock.py
@@ -53,6 +53,7 @@ launcher in any JupyterLab host, CI, the CLI) gets it.
 from __future__ import absolute_import
 
 from anaconda_project.internal import conda_api
+from anaconda_project.internal.py2_compat import is_string, is_dict
 
 
 # anaconda-project lock files group package specs into "buckets" keyed by
@@ -116,19 +117,52 @@ def parse_locked_specs(lock_set, platform):
         raise LockTranslationError("lock set has no entry for platform %r (has %r)" %
                                    (platform, list(lock_set.platforms)))
 
+    if lock_set.disabled:
+        raise LockTranslationError("lock set for platform %r is not locked (locking disabled)" % platform)
+
+    try:
+        specs_for_platform = lock_set.package_specs_for_platform(platform)
+    except TypeError as e:
+        # package_specs_for_platform's own bucket-merge (_conda_combine_key)
+        # calls conda_api.parse_spec on every entry BEFORE we see any of
+        # them, so a non-string entry (e.g. a nested {"pip": [...]} bucket
+        # parsed alongside conda specs) raises conda_api's raw TypeError
+        # here, not in the loop below. assert_no_pip_packages only catches
+        # the top-level "pip" platform key; surface this case as our own
+        # typed error too, rather than let the TypeError leak.
+        raise LockTranslationError(
+            "locked spec list for platform %r contains a non-spec-string entry: %s" % (platform, e))
+
     locked = []
-    for spec in lock_set.package_specs_for_platform(platform):
+    for spec in specs_for_platform:
+        if not is_string(spec):
+            # Belt-and-suspenders: a non-string entry that survives the
+            # bucket merge above (e.g. from a single-bucket platform with no
+            # merge to trigger the TypeError) is caught here too.
+            raise LockTranslationError(
+                "locked spec for platform %r is not a package spec string: %r" % (platform, spec))
         parsed = conda_api.parse_spec(spec)
         if parsed is None:
             raise LockTranslationError("could not parse locked spec %r for platform %r" % (spec, platform))
-        if parsed.exact_version is None or parsed.exact_build_string is None:
-            # A real lock pins name=version=build. A spec missing the version
-            # or build means this platform isn't actually locked -> a re-solve
-            # would be required, which is precisely what we refuse to do.
+        # conda_api.parse_spec's exact_version rejects |,* -- but that filter
+        # does NOT extend to exact_build_string, so a build of e.g. "*" or
+        # "py37*" is (wrongly) reported as "exact". A wildcard/partial build
+        # is exactly the "this platform was never truly locked" condition
+        # this strict gate exists to catch -- reject it explicitly rather
+        # than let it surface later as a misattributed BuildNotFoundError.
+        build = parsed.exact_build_string
+        if (parsed.exact_version is None or build is None or
+                any(c in build for c in ('*', '|', ','))):
             raise LockTranslationError(
                 "locked spec %r for platform %r is not pinned to an exact version=build; "
                 "the lock set is not fully resolved for this platform" % (spec, platform))
-        locked.append((parsed.name, parsed.exact_version, parsed.exact_build_string))
+        locked.append((parsed.name, parsed.exact_version, build))
+
+    names = [name for (name, _version, _build) in locked]
+    if len(names) != len(set(names)):
+        dupes = sorted({n for n in names if names.count(n) > 1})
+        raise LockTranslationError(
+            "locked spec list for platform %r has duplicate/conflicting pins for: %s" % (platform, ", ".join(dupes)))
     return locked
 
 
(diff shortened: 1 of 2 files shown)


mcg1969 (approved):
## Review: no-solver `anaconda-project-lock.yml` → `pixi.lock` translator

**Reviewed and approved.** The design is sound, the code is careful, and I verified the central claim end-to-end against real repodata and a real `pixi` — including the `pixi install --locked` acceptance gate this PR lists as still pending.

### What I did

Ran the suite, then built real E2E cases the offline tests can't cover.

**Unit suite:** 30/30 pass, including the live repodata canary (conda 26.3.2). Only additive changes — two new files, nothing modified or deleted, so no regression surface.

**E2E case 1 — single-platform, real repodata → real pixi (the pending acceptance gate):**

* Created a real project (`python=3.12`, `zlib`), ran `anaconda-project lock` → a genuine 19-package pinned lock.
* Loaded it as a real `CondaLockSet`, ran `translate_lock_set` against live `repo.anaconda.com/pkgs/main` → real URLs, sha256, md5, depends. noarch packages (pip, tzdata) correctly emitted under `/noarch/` URLs.
* **`pixi install --locked` accepted the translated lock and installed all 19 at their exact locked builds — no re-solve.** This is the item the PR description flags as pending; it passes.

**E2E case 2 — multi-platform (foreign-subdir + cross-platform noarch dedup):**

* linux-64 + osx-arm64 lock, translated on an osx-arm64 host → foreign linux-64 records enriched correctly; shared noarch packages deduped (48 top-level = 31 + 19 − 2 shared). `pixi install --locked` accepted it.

**Fail-loud paths, exercised live:**

* Tampered build → `BuildNotFoundError` (channel answered, build absent — no substitution).
* Unreachable host → `MirrorUnavailableError`, *not* misclassified as build-gone.
* Hand-broke the closure (dropped a package) → pixi rejects with exactly `lock-file not up-to-date`, validating the atomic-write / full-closure rationale.

**Convention check:** a real pixi-generated lock references noarch packages by their `/noarch/` URL from within each platform's list — precisely what the translator emits. Schema match confirmed against a real lock, not assumed.

### Findings

No correctness defects. Two minor, non-blocking observations:

* **Duplicate-pin guard is intra-bucket only.** Same name in `all` + `<platform>` resolves via `CondaLockSet`'s own precedence (platform wins), with no error. This is correct — bucket precedence is anaconda-project's to own — but a one-line comment near the dup check would keep a future reader from mistaking it for a gap.
* **`enrich_locked_packages` catches bare `Exception` as a transport failure** (pixi_lock.py around line 312). A genuine bug inside `query` (rather than a transport problem) would surface as `MirrorUnavailableError`. Fine for v1 — the CI canary guards the API shape and misclassifying as "retry the mirror" is low-harm — but it's the one spot that could mislead debugging.

The consumer wiring (calling `translate_lock_set` from the convert path) remains a real follow-up, as noted. This PR is the library, and the library is correct.

PR #440, opened 2026-07-02
Add conda-workspaces support (conda.toml read/write, pyproject.toml embedding)

mcg1969:
## Summary

* Adds a third project manifest type/format, **conda-workspaces** (`conda.toml`, plus `pyproject.toml` embedding via `[tool.conda]`/`[tool.pixi]`), on both the read side (`project_info.publication_info`) and a new export side (`export-conda` CLI command).
* Reorders manifest-detection precedence to `conda.toml` > `pixi.toml` > `anaconda-project.*` > `pyproject.toml` (previously just `pixi.toml` > `anaconda-project.yml`), matching conda-workspaces' own documented directory-scan precedence.
* Renames `docs/source/pixi-support.rst` → `docs/source/workspace-support.rst` to cover both export targets.

## How this was built

This went through a full plan → spec → execute → review → fix loop:

* The design was ground-truthed against the real, installed `conda-workspaces` 0.7.0 Python package (not just its docs) — confirming its TOML parser's `args` shorthand acceptance, that `[workspace]`'s `additionalProperties: false` schema constraint isn't actually enforced at runtime, the real nesting of `pyproject.toml` embedding (`[tool.conda.workspace]`, not a flat `[tool.conda]`), and that `conda.lock` shares `pixi.lock`'s `version`/`environments`/`packages` shape (stamped `version: 1` instead of `6`).
* A first pass landed 8 gated commits implementing the feature.
* A 7-persona review standup (code-reviewer, python-reviewer, silent-failure-hunter, sre, staff-engineer, product-manager, devils-advocate) then reviewed the diff and found 2 confirmed, independently-reproduced bugs plus a confirmed API asymmetry. All three were independently re-verified against the real code before fixing.
* A second pass (7 more commits) fixed every confirmed finding, collapsed two duplication hotspots the reviewers flagged, and — caught during final independent verification, not by the review itself — fixed a wording regression the duplication-collapse had silently introduced, with a new test locking in the correct wording for both formats.

## What changed, in commit order

1. `conda.toml` and `pyproject.toml` detection in `detect_project_type()`
2. Refactor `_pixi_publication_info` to accept a pre-parsed data dict (enables format-agnostic reuse)
3. `conda-workspaces` `publication_info` branch + `pyproject.toml` fallback for both formats
4. `anaconda-project info` CLI surfaces the `conda-workspaces` project type
5. `export-conda` `project_ops` functions (`export_conda`, `preview_conda_export`)
6. `export_conda_toml` writer (delegates to `export_pixi_toml`, since the two TOML shapes are field-for-field identical for the fields this exporter emits)
7. `export-conda` CLI wiring (new `export-conda` subcommand)
8. Docs retitle: `workspace-support.rst` covers both pixi and conda-workspaces
9. **Fix**: stop leaking pixi-specific prose into `conda.toml` output (e.g. `# default (pixi creates this implicitly...)`, `...before running pixi` — the exporter's "zero content differences" claim was false for projects with services or multiple env_specs)
10. **Fix**: `[tool.anaconda.commands.*]` overrides were silently dropped for `pyproject.toml`-embedded manifests (wrong lookup path — was looking for `tool.pixi.tool.anaconda`/`tool.conda.tool.anaconda` instead of the document's top-level `tool.anaconda`)
11. **Fix**: `--project-type` explicit override now recognizes `pyproject.toml` embedding, matching auto-detect (previously `anaconda-project info --project-type pixi` failed on a directory that `anaconda-project info` alone correctly identified as a pixi project)
12. Collapse `_pixi_publication_info`/`_conda_workspaces_publication_info` (~90% duplicate logic) into one shared, parameterized implementation
13. Collapse `conda_export_commands.py`/`pixi_commands.py` (near-identical CLI wiring) into one shared `_export_workspace` helper
14. Small hygiene fixes: unused import, stale docstrings, narrowed exception handling in lockfile readers
15. **Fix**: restore "Pixi" capitalization in `export-pixi`'s CLI recommendation text, which the duplication collapse (item 13) had silently flattened to lowercase — plus a new test asserting the exact wording for both formats so this can't regress silently again

## Explicit non-goals (scoped out, on purpose)

* No `pyproject.toml` *writer* — reads only.
* No `api.py` changes — `project_ops`/`project_info` stay the real public surface, matching the pre-existing state for `export_pixi`/`preview_pixi_export`.
* No `CondaExportStatus` class — `export_conda` reuses `PixiExportStatus` (its two fields are generic export concepts, not pixi-specific).
* No support for conda.toml-only export fields (`[workspace.archive]`, `channel-priority`, `envs-dir`, etc.) — `anaconda-project.yml` has no source concept for any of them.

## Test plan

* [x] Full targeted gate green: `python -m pytest anaconda_project/test/test_project_info.py anaconda_project/internal/test/test_pixi_export.py anaconda_project/internal/cli/test/test_info.py anaconda_project/internal/test/test_conda_export.py -q` — 201 passed on the pre-change baseline, 247 passed after this PR (46 new tests added across both passes).
* [x] Full repo suite: 1265 passed, 10 skipped, 1 failed. The 1 failure (`test_prepare.py::test_default_to_system_environ`) is pre-existing and environment-specific (confirmed by reproducing the identical failure against the unmodified pre-change commit in an isolated worktree) — unrelated to this change.
* [x] Manually verified against the real, installed `conda-workspaces` 0.7.0 package: `conda workspace info` / `conda task list` against hand-written `conda.toml` fixtures, and direct calls into `conda_workspaces.manifests.toml.CondaTomlParser`/`normalize_args` to confirm shape assumptions before writing the reader/exporter.
* [x] Manually reproduced and re-verified all 3 fixed bugs (prose leak, dropped command override, `--project-type` asymmetry) directly against this branch's code, independent of the automated test suite, before and after each fix.

### DIFF
--- a/anaconda_project/internal/cli/conda_export_commands.py
+++ b/anaconda_project/internal/cli/conda_export_commands.py
@@ -8,61 +8,14 @@
 """Commands related to conda-workspaces (conda.toml) export."""
 from __future__ import absolute_import, print_function
 
-from anaconda_project.internal.cli.project_load import load_project
-from anaconda_project.internal.cli import console_utils
+from anaconda_project.internal.cli.pixi_commands import _export_workspace
 from anaconda_project import project_ops
 
 
 def export_conda(project_dir, filename, use_default=False, add_current_platform=False):
     """Export the project as a conda.toml file."""
-    import os
-    project = load_project(project_dir)
-    if console_utils.print_project_problems(project):
-        return 1
-    # Default to writing conda.toml in the project directory
-    if filename == 'conda.toml' and not os.path.isabs(filename):
-        filename = os.path.join(project_dir, filename)
-    status = project_ops.export_conda(project, filename=filename,
-                                      use_default=use_default,
-                                      add_current_platform=add_current_platform)
-    if status:
-        print(status.status_description)
-        # When the corresponding flag actually changed something, the
-        # status carries the value (rename source / added platform);
-        # surface it so the user can see what we did. When the user
-        # didn't pass the flag but the project would benefit, recommend
-        # the flag instead.
-        from anaconda_project.internal.pixi_export import (
-            current_platform_addition_target, default_rename_target,
-        )
-        if use_default and status.default_rename_from:
-            print(
-                'The "{}" environment has been renamed "default" in the '
-                'exported conda-workspaces specification.'.format(status.default_rename_from))
-        elif not use_default:
-            promoted = default_rename_target(project)
-            if promoted is not None:
-                print(
-                    'Recommendation: re-run using the --use-default flag to '
-                    'rename the "{}" environment to "default". This will '
-                    'simplify the resulting conda-workspaces specification.'.format(promoted))
-        if add_current_platform and status.current_platform_added:
-            print(
-                'The current platform "{}" has been added to the platforms '
-                'list in the exported conda-workspaces specification.'
-                .format(status.current_platform_added))
-        elif not add_current_platform:
-            missing = current_platform_addition_target(project)
-            if missing is not None:
-                print(
-                    'Recommendation: re-run using the --add-current-platform '
-                    'flag to add "{}" to the platforms list. conda-workspaces '
-                    'requires the host platform to be declared before it will '
-                    'install the environment.'.format(missing))
-        return 0
-    else:
-        console_utils.print_status_errors(status)
-        return 1
+    return _export_workspace(project_dir, filename, use_default, add_current_platform,
+                            project_ops.export_conda, 'conda.toml', 'conda-workspaces')
 
 
 def main_export_conda(args):
--- a/anaconda_project/internal/cli/conda_export_commands.py
+++ b/anaconda_project/internal/cli/conda_export_commands.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2016, Anaconda, Inc. All rights reserved.
+#
+# Licensed under the terms of the BSD 3-Clause License.
+# The full license is in the file LICENSE.txt, distributed with this software.
+# -----------------------------------------------------------------------------
+"""Commands related to conda-workspaces (conda.toml) export."""
+from __future__ import absolute_import, print_function
+
+from anaconda_project.internal.cli.project_load import load_project
+from anaconda_project.internal.cli import console_utils
+from anaconda_project import project_ops
+
+
+def export_conda(project_dir, filename, use_default=False, add_current_platform=False):
+    """Export the project as a conda.toml file."""
+    import os
+    project = load_project(project_dir)
+    if console_utils.print_project_problems(project):
+        return 1
+    # Default to writing conda.toml in the project directory
+    if filename == 'conda.toml' and not os.path.isabs(filename):
+        filename = os.path.join(project_dir, filename)
+    status = project_ops.export_conda(project, filename=filename,
+                                      use_default=use_default,
+                                      add_current_platform=add_current_platform)
+    if status:
+        print(status.status_description)
+        # When the corresponding flag actually changed something, the
+        # status carries the value (rename source / added platform);
+        # surface it so the user can see what we did. When the user
+        # didn't pass the flag but the project would benefit, recommend
+        # the flag instead.
+        from anaconda_project.internal.pixi_export import (
+            current_platform_addition_target, default_rename_target,
+        )
+        if use_default and status.default_rename_from:
+            print(
+                'The "{}" environment has been renamed "default" in the '
+                'exported conda-workspaces specification.'.format(status.default_rename_from))
+        elif not use_default:
+            promoted = default_rename_target(project)
+            if promoted is not None:
+                print(
+                    'Recommendation: re-run using the --use-default flag to '
+                    'rename the "{}" environment to "default". This will '
+                    'simplify the resulting conda-workspaces specification.'.format(promoted))
+        if add_current_platform and status.current_platform_added:
+            print(
+                'The current platform "{}" has been added to the platforms '
+                'list in the exported conda-workspaces specification.'
+                .format(status.current_platform_added))
+        elif not add_current_platform:
+            missing = current_platform_addition_target(project)
+            if missing is not None:
+                print(
+                    'Recommendation: re-run using the --add-current-platform '
+                    'flag to add "{}" to the platforms list. conda-workspaces '
+                    'requires the host platform to be declared before it will '
+                    'install the environment.'.format(missing))
+        return 0
+    else:
+        console_utils.print_status_errors(status)
+        return 1
+
+
+def main_export_conda(args):
+    """Start the export-conda command and return exit status code."""
+    return export_conda(args.directory, args.filename,
+                        use_default=args.use_default,
+                        add_current_platform=args.add_current_platform)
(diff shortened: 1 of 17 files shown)


A-Aron-T:
## Blocking finding (with a tested fix ready to merge): explicit `--project-type` override falsifies its own documented "mid-conversion" contract

Ran a full review pass (independent code review + live reproduction in a real editor session container + a 7-persona pocket-team standup, 5 personas independently corroborating on the same root cause, 2 at blocker severity). One real bug, confirmed three ways — and a fix, also live-verified, waiting in #441.

### The bug

`publication_info()`'s explicit `project_type=` override (and the `--project-type` CLI flag) still routes through `_locate_manifest()`, which unconditionally runs the *full* `conda.toml > pixi.toml > anaconda-project.* > pyproject.toml` precedence scan and only accepts the single highest-precedence match. If a directory has **both** `conda.toml` and `pixi.toml` — the exact "mid-conversion" scenario `--project-type` exists to support, per `docs/source/workspace-support.rst:301` — requesting `project_type='pixi'` raises `FileNotFoundError: No pixi.toml found in <dir>` even though `pixi.toml` demonstrably exists on disk.

Live-reproduced via the actual CLI in a real editor-lite container (same image DSW sessions boot from):
```
$ anaconda-project info --project-type pixi --directory /tmp/fixture
No pixi.toml found in /tmp/fixture (and no [tool.pixi.workspace] in pyproject.toml)
$ echo $?
1
```
...with `pixi.toml` verifiably present in that directory. Same failure mode the other direction — a top-level `pixi.toml` masking a `conda.toml` that only exists as a `pyproject.toml` embedding (`[tool.conda.workspace]`) — also live-reproduced with `project_type='conda-workspaces'`.

**`project_type='anaconda-project'` is unaffected** — it does its own direct `os.path.isfile()` loop over `possible_project_file_names`, bypassing `_locate_manifest` entirely. That's actually the correct pattern; `pixi`/`conda-workspaces` just don't follow it.

Zero test coverage exists for "both manifests present + override requests the non-winning one" in either direction — every existing override test (including this PR's new ones) only exercises the requested format when it's the sole manifest present or already the precedence winner.

### Fix — ready to merge: #441

Opened **#441** (`fix/aent-explicit-project-type-precedence-mask` → `feature/conda-workspaces-support`, i.e. targets *this* branch, not `master`) with a tested fix: mirrors the already-correct `anaconda-project` branch by adding `_locate_manifest_of_kind()`, a direct existence check for the one requested format (top-level file, then pyproject embedding) that never delegates to `_locate_manifest`'s cross-format precedence scan. Used in both the `pixi` and `conda-workspaces` explicit-override branches.

Includes 4 new regression tests — both directions of "both manifests present, request the loser" (top-level vs. top-level, and each direction's pyproject-embedded variant). Confirmed 3 of them fail without the fix and all pass with it. Full targeted gate: **251/251 passing** (247 baseline + 4 new).

**Live-verified in a real `ae-editor-lite` container** (the actual DSW session image), via the real CLI:
```
$ anaconda-project info --project-type pixi --directory /tmp/fixture   # both conda.toml + pixi.toml present
# before #441: exit 1, FileNotFoundError, despite pixi.toml existing on disk
# after #441:  exit 0, correct pixi data returned
```
No regression in the reverse direction, the single-manifest happy path, or the correctly-still-rejected truly-absent case.

Since #441 targets this branch (not `master`), CI won't auto-trigger on it — the repo's workflow only runs for PRs into `master` — but it's already exercised against your branch's exact code locally and in the live container above. Should be a straightforward merge into `feature/conda-workspaces-support`.

### Second, separate doc inaccuracy (not just "undersells the bug")

`docs/source/workspace-support.rst:299-304` and `:344-353` state that forcing `--project-type` "only recognizes the corresponding top-level manifest... not a pyproject.toml embedding." That's factually wrong about a part of the code that's *actually correct* — `project_info.py`'s own docstring and the passing `test_explicit_pixi_recognizes_pyproject_pixi`/`test_explicit_conda_recognizes_pyproject_conda` tests confirm the override **does** recognize pyproject.toml embeddings. Worth fixing in the same docs pass as the mid-conversion caveat, so the page doesn't end up with a third, still-incomplete description.

### Non-blocking, raised for awareness

The precedence reorder (`conda.toml > pixi.toml > anaconda-project.* > pyproject.toml`) is a real behavior change for any existing `anaconda-project.yml`/`pixi.toml` user who happens to have an unrelated `conda.toml` in the same directory (stray file, template scaffold, partial migration) — it's riding along inside a "new format" PR. Not asking to hold on this, just flagging it as a distinct risk surface from the override bug, worth a second look/explicit sign-off given it changes auto-detection for existing users.

---
Happy to help with the fix if useful. Everything above was independently reproduced against a live checkout of this branch, not just read from the diff.

mcg1969:
Thanks for the thorough review — all three items are now resolved. Point-by-point:

**1. Blocking bug (explicit `--project-type` override masked by precedence) — fixed and merged.**
#441 is merged into this branch (merge commit `015e69c`). I independently re-verified before merging: applied the 4 new tests on top of the pre-fix `project_info.py` — 3 fail, 1 passes (the passing one is the reverse direction that already worked, `conda.toml` outranking `pixi.toml`, so it's a lock-in). With the fix all 4 pass, and the targeted gate is green at 251. The fix (`_locate_manifest_of_kind()`) mirrors the already-correct `anaconda-project` branch's direct existence check, so a coexisting higher-precedence manifest of a different format no longer masks the requested one.

**2. Doc inaccuracy (`--project-type` "only recognizes the top-level manifest, not a pyproject.toml embedding") — fixed and merged in #442.**
You were right, and #441 made it demonstrably wrong. #442 (docs-only, merged into this branch as `7b72f62`) rewrites both passages — the CLI-flag docs and the `publication_info` optional-args list — to state that the override matches the requested format in *either* form (top-level or `pyproject.toml` embedding), independent of precedence, erroring only when it's present in neither. #441's own `test_force_pixi_via_pyproject_when_conda_toml_also_present` is the proof the old wording was false.

**3. Precedence reorder — signing off; we reject the "stray file" premise.**
A `conda.toml` is not an accidental stray file that happens to be lying around. Its presence signals a deliberate migration to conda-workspaces, exactly as a `pixi.toml` signals a move off `anaconda-project`. A higher-precedence `conda.toml` winning auto-detection is the correct, intended behavior — not a risk surface. No change here.

With #441 and #442 both merged into this branch, the review feedback is fully addressed.

A-Aron-T (approved):
## Verified all 3 items from the earlier blocking review are resolved

Cloned `feature/conda-workspaces-support` and independently confirmed:

- **Full targeted test gate: 251/251 passing** (`test_project_info.py` + `test_info.py` + `test_pixi_export.py` + `test_conda_export.py`), matching the claimed count exactly.
- **Live-reproduced the original bug scenario is fixed**: a fixture with both `conda.toml` and `pixi.toml` present. `anaconda-project info --project-type pixi` now returns `Type: pixi` (exit 0) instead of the previously reported `FileNotFoundError`. Confirmed the reverse direction too (`--project-type conda-workspaces` → `Type: conda-workspaces`, exit 0).
- **Auto-detect precedence confirmed**: same fixture with no override correctly resolves to `conda-workspaces` per the new `conda.toml > pixi.toml > anaconda-project.* > pyproject.toml` order.
- `_locate_manifest_of_kind()` is live in the branch (`project_info.py`), wired into both the `pixi` and `conda-workspaces` explicit-override branches as described.

CI green across the full OS/Python matrix. No further changes needed — approving.

PR #441, opened 2026-07-14
fix(project_info): explicit --project-type no longer masked by manifest precedence

A-Aron-T:
## Summary

Targets #440's branch directly (not `master`) — this fixes a bug found while reviewing that PR, scoped to land alongside it rather than as a separate follow-up.

`publication_info()`'s explicit `project_type='pixi'`/`'conda-workspaces'` override delegates to `_locate_manifest()`, which unconditionally runs the *full* `conda.toml > pixi.toml > anaconda-project.* > pyproject.toml` precedence scan and only accepts the single highest-precedence match. If a directory has **both** `conda.toml` and `pixi.toml` — the "mid-conversion" scenario `--project-type` exists to support, per `workspace-support.rst` — requesting `project_type='pixi'` raises `FileNotFoundError` even though `pixi.toml` demonstrably exists on disk.

`project_type='anaconda-project'` was already unaffected — it does its own direct `os.path.isfile()` loop, bypassing `_locate_manifest` entirely.

## Fix

Adds `_locate_manifest_of_kind(project_dir, top_level_filename, pyproject_tool_key)`, which checks only whether the *one requested* format's manifest exists (top-level file, or a `pyproject.toml` `[tool.<key>.workspace]` table) — never delegating to the cross-format precedence scan. Used in both the `conda-workspaces` and `pixi` explicit-override branches, mirroring the already-correct `anaconda-project` branch's direct-check pattern.

## Tests

3 new regression tests covering both directions of "both manifests present, explicit override requests the non-winning one," plus the pyproject-embedded variant. Confirmed 2 of the 3 fail without the fix (`git stash` A/B on `project_info.py` only, tests kept) and all pass with it.

Full targeted gate (`test_project_info` + `test_pixi_export` + `test_info` + `test_conda_export`): **250 passed** (247 baseline + 3 new), matching #440's own claimed test count.

**Live-verified in a real `ae-editor-lite` container** (the actual DSW session image), via the real `anaconda-project info --project-type pixi` CLI:
- Before this fix: `exit 1`, `FileNotFoundError`, against a directory where `pixi.toml` demonstrably exists.
- After: `exit 0`, correct `pixi` data returned.
- Confirmed no regression in the reverse direction (`conda-workspaces` requested when `pixi.toml` also exists), the single-manifest happy path, or the correctly-still-rejected truly-absent case.

### DIFF
--- a/anaconda_project/project_info.py
+++ b/anaconda_project/project_info.py
@@ -97,6 +97,40 @@ def _locate_manifest(project_dir):
     return (None, None)
 
 
+def _locate_manifest_of_kind(project_dir, top_level_filename, pyproject_tool_key):
+    """Find the manifest for one specific format, ignoring precedence.
+
+    Unlike :func:`_locate_manifest` (which returns the single highest-precedence
+    manifest across *all* formats), this checks only whether *this* format's
+    manifest exists — a top-level file named *top_level_filename*, or a
+    ``pyproject.toml`` with a ``[tool.<pyproject_tool_key>.workspace]`` table.
+    This is what explicit ``project_type=`` overrides need: "does the requested
+    format exist here?", not "which format wins?" — so a coexisting
+    higher-precedence manifest of a *different* format never masks the one the
+    caller actually asked for.
+
+    Returns (manifest_kind, path) — manifest_kind is ``'top-level'`` or
+    ``'pyproject'`` — or (None, None) if not found. Parse errors in
+    pyproject.toml are silently treated as non-match (returns (None, None)),
+    matching _locate_manifest's existing convention.
+    """
+    top_level_path = os.path.join(project_dir, top_level_filename)
+    if os.path.isfile(top_level_path):
+        return ('top-level', top_level_path)
+
+    pyproject_path = os.path.join(project_dir, PYPROJECT_MANIFEST)
+    if os.path.isfile(pyproject_path):
+        try:
+            with open(pyproject_path, 'rb') as f:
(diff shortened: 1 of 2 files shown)


mcg1969:
Verified and merging.

Independently reproduced the fix against this branch:

- **A/B confirmed the tests catch the bug:** applied the 4 new tests on top of the pre-fix `project_info.py` — 3 fail, 1 passes. The one that passes is exactly the reverse direction that already worked (`conda.toml` outranks `pixi.toml`), so it's a lock-in, not a regression test. With the fix, all 4 pass.
- **Targeted gate green:** 251 passed (`test_project_info` + `test_pixi_export` + `test_info` + `test_conda_export`).
- **Fix is the right shape:** `_locate_manifest_of_kind()` mirrors the already-correct `anaconda-project` branch's direct existence check — top-level file, then `[tool.<key>.workspace]` embedding — so a coexisting higher-precedence manifest of a *different* format no longer masks the one the caller explicitly asked for.
- Clean merge, based on the current tip of `feature/conda-workspaces-support`.

Merging into `feature/conda-workspaces-support` as a merge commit (preserving both commits).

One follow-up this PR intentionally scoped out: `docs/source/workspace-support.rst` still claims `--project-type` "only recognizes the corresponding top-level manifest ... not a `pyproject.toml` embedding." That's now disproved by this PR's own `test_force_pixi_via_pyproject_when_conda_toml_also_present`. Opening a separate docs PR into the feature branch to correct it.

